PR #17, opened 2014-10-26
[WIP] Add report-uri

### LINKED ISSUE #12: Add report-uri http://localhost:4200
Chrome throws additional errors when the CSP is active in Report-only mode but no report-uri is configured. 

I think adding `report-uri http://localhost:4200` to the CSP header would be fine for now.

Later on, we could add a blueprint for the server where we actually process the CSP reports.

joostdevries:
Fixes #12 
- Adds a `report-uri` header which defaults to host:port.
- Now just silently returns `{"status":"ok"}` to all requests on `/csp-report`

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,6 +41,11 @@ module.exports = {
         });
       }
 
+      if (header.indexOf('Report-Only')!==-1 && !('report-uri' in headerConfig)) {
+        headerConfig['connect-src'] = headerConfig['connect-src'] + ' http://' + options.host +':' + options.port + '/csp-report';
+        headerConfig['report-uri'] = 'http://' + options.host +':' + options.port + '/csp-report'; 
+      }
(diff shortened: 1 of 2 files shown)


mgenev:
strangely, I'm still getting the 'report-only' issue with ember-cli 0.1.7

agrobbin:
@mgenev looks like this change was never released as a new version. @rwjblue could we get a new release of this?

EmergentBehavior:
Experiencing this still as well on v0.1.12.

HeroicEric:
Fo anybody else still experiencing this, looks like change still hasn't been released (as of v0.3.0).

shaunc:
Its in ember-cli-content-security-policy=0.4.0 -- you may have to set in app package.json explicitly.

PR #34, opened 2015-04-24
Upgrade to ember-cli 0.2.3

### LINKED ISSUE #32: Build hangs under iojs
It looks like this code doesn't build under any version of iojs; running `ember build` causes the process to hang (and peg the CPU depending on the version of iojs).  I started with the latest (1.8.1) and then cherry picked my way back to 1.0.0, all with the same result.  I used nvm so I was always getting the correct iojs/npm pair.  I'm running on OS X 10.10.3.

This is totally weird and I've never seen this sort of thing before. The build works fine under node.js 0.12.2.

ZebraFlesh:
Upgrade to ember-cli 0.2.3 and latest body-parser (1.12.3). Removed unnecessary ember-data dependency. Added MIT license.
Fixes #32


rwjblue:
Thanks!

PR #35, opened 2015-04-24
Log CSP reports from Firefox

### LINKED ISSUE #31: Reporter not logging reports from Firefox
The development reporter only accepts POSTs of type [`application/csp-report`](https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L71).  Unfortunately it looks like the latest version of Firefox is doing POSTs with a mime type of `application/json`. The net result is that Firefox doesn't spit out any console logging for CSP violations, making it difficult to debug.  Digging into CSP, it looks like Firefox correctly flags a lot of things that other browsers don't (specifically around SVG).

ZebraFlesh:
Firefox sends CSP reports with mime type application/json. Remove unused variable.
Fixes #31

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@ module.exports = {
         'style-src': "'self'",
         'media-src': "'self'"
       }
-    }
+    };
 
     if (environment === 'development') {
       ENV.contentSecurityPolicy['script-src'] = ENV.contentSecurityPolicy['script-src'] + " 'unsafe-eval'";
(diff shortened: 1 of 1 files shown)


AW-UC:
I patched `index.js` according to this, but I don't see any change in logging in latest stable FF on Windows. Anything I missed?

ZebraFlesh:
Refer to #36: the report mechanism doesn't appear to work on windows.

rwjblue:
Thank you!

PR #37, opened 2015-04-29
Normalize options.host so report-uri works on windows

### LINKED ISSUE #36: CSP reporter does not work on windows
During normal local development, the report-uri is computed as http://0.0.0.0:4200/csp-report.  On unix based systems (including OS X), this is not a problem and the 0.0.0.0 is properly resolved to localhost or some equivalent.  On Windows, the POST fails completely and CSP report is never logged to the console.  Windows just doesn't like this address.

ZebraFlesh:
Fixes #36
Didn't touch the live reload host as it looks like that's hard coded to localhost and 0.0.0.0 for some reason.  Should that be fixed in the future?

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,7 @@ module.exports = {
 
       var header = appConfig.contentSecurityPolicyHeader;
       var headerConfig = appConfig.contentSecurityPolicy;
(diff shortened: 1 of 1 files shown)


rwjblue:
LGTM

PR #41, opened 2015-05-08
Inject http-equiv metatag into head

theodorton:
As seen in [this issue][1], as of iOS 7, you may provide Content-Security-Policy in a metatag. This makes it possible to enable security measures provided by CSP in environments where headers simply aren't present (like phonegap).

I'm using the `addon.contentFor` hook to inject the meta tag. The content should be more or less equivalent to what is served from `ember serve`, with the exception that you'll have to explicitly provide the `report-uri` directive (though I'm not sure if this directive will be used in a local environment).

I don't see this PR as ready to merge, but I would like some input.

What should be done:
- [ ] Refactor `_headerData` with a better argument list
- [x] Make the meta tag opt-in with some configuration variable

[1]: https://github.com/MobileChromeApps/mobile-chrome-apps/issues/73

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,40 @@
+var _headerData = function(config) {
+  var options = config.options;
+  var project = options.project;
+
+  var appConfig = project.config(options.environment);
+
+
+  var header = appConfig.contentSecurityPolicyHeader;
+  var headerConfig = appConfig.contentSecurityPolicy;
+  var normalizedHost = options.host === '0.0.0.0' ? 'localhost' : options.host;
+
+  if (options.liveReload) {
+    ['localhost', '0.0.0.0'].forEach(function(host) {
+      headerConfig['connect-src'] = headerConfig['connect-src'] + ' ws://' + host + ':' + options.liveReloadPort;
+      headerConfig['script-src'] = headerConfig['script-src'] + ' ' + host + ':' + options.liveReloadPort;
+    });
+  }
+
+  if (header.indexOf('Report-Only')!==-1 && !('report-uri' in headerConfig)) {
+    headerConfig['connect-src'] = headerConfig['connect-src'] + ' http://' + normalizedHost + ':' + options.port + '/csp-report';
+    headerConfig['report-uri'] = 'http://' + normalizedHost + ':' + options.port + '/csp-report';
+  }
+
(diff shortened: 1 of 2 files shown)


theodorton:
Any comments?

devinus:
Wow, I was just looking for this. I need to use this because we use Cloudfront/S3 and it doesn't support setting CSP headers. What can we do to make this happen?

sandstrom:
Awesome! :smile: I was just about to open a PR for this.

@rwjblue what are your thoughts on this? I find delivery via meta much more useful than http-header, because of better workflow (server-configuration is often separate from application development).

I have implemented something similar in our project (copied below for reference):

```js
contentFor: function(type, config) {
    if (type === 'head') {
      var policyObject = config.contentSecurityPolicy;

      var policyString = Object.keys(policyObject).reduce(function(memo, value) {
        if (value === null || value === undefined) {
          return memo;
        } else {
          return memo + value + ' ' + policyObject[value] + '; ';
        }
      }, '');

      return '<meta http-equiv="Content-Security-Policy" content="' + policyString + '">';
    }
  }
```

alexgb:
I believe the CSP meta tag doesn't support all security features. Would at least be worth mentioning this in the Readme. Otherwise :+1:, this would be helpful in my use case.

sandstrom:
@alexgb All except `frame-ancestors` and `sandbox` are supported with the `<meta ...` tag.

index.js
```diff
@@ -1,3 +1,38 @@
+var _headerData = function(appConfig, options) {
```
rwjblue: Can you move this into a utility function?  I would like to share this with a future command that will emit the header value to the console...

rwjblue:
I'm :+1: on this.

@theodorton - Mind rebasing?

sandstrom:
@rwjblue awesome, I really like this plugin!

topaxi:
Firefox Nightly (45) now also supports &lt;meta&gt; based CSP -> https://bugzilla.mozilla.org/show_bug.cgi?id=663570

rwjblue:
Needs a rebase, and tweaks identified during review (mentioned above).

sandstrom:
Should be solved now (https://github.com/rwjblue/ember-cli-content-security-policy/issues/54), I think this can be closed. Please try out the code in master! :)

PR #42, opened 2015-05-28
Improved console output

Turbo87:
see commits


Turbo87:
@rwjblue any comments?

Turbo87:
@rwjblue anyone home? some form of response would be much appreciated.

rwjblue:
Seems good to me, thanks for being patient!

PR #44, opened 2015-06-12
Use https if options.ssl is true for CSP report uri

### LINKED ISSUE #43: csp-report not working on https
When running ```ember serve``` with the ```--ssl``` the csp report is using http instead of https and the request is blocked by the browser.

jrobeson:
This fixes #43

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,8 +42,9 @@ module.exports = {
       }
 
       if (header.indexOf('Report-Only')!==-1 && !('report-uri' in headerConfig)) {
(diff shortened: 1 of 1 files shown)


ezpuzz:
would be nice if it also handled `wss://` vs `ws://` for liveReload.

sandstrom:
This is fixed on master. Can be solved https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L88

PR #45, opened 2015-06-14
Use options.liveReloadHost for LiveReload CSP

### LINKED ISSUE #18: Different development hostname
I use a hostname, something like:

`http://portfolio.dev`

for some of my projects. Could we make https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L38 work with a config'ed var instead of hard-coding those 2 options?

I'll take a stab at it if I get some time.

jrobeson:
`options.liveReloadHost` has been implemented in https://github.com/ember-cli/ember-cli/pull/4292.

This should fix #18.

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,9 +35,11 @@ module.exports = {
       var normalizedHost = options.host === '0.0.0.0' ? 'localhost' : options.host;
 
       if (options.liveReload) {
-        ['localhost', '0.0.0.0'].forEach(function(host) {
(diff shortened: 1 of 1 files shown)


max:
Thanks for doing this. The hardcoded host bit me too. Hopefully we can get this merged soon.

sandstrom:
@jrobeson Thanks, much appreciated! ⛵ 

This should be solved in master (https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L112-L116). I'll go ahead and close this, but please let me know if you think this should be reopened.

PR #49, opened 2015-09-21
Update README.md

bttf:
Mention in README.md that plugin in already included in ember-cli


sandstrom:
This can be closed since it's no longer included by default. https://github.com/ember-cli/ember-cli/pull/5247

PR #39, opened 2015-05-03
Fix development config

ZebraFlesh:
Moved development config out of addon configuration path and into express middleware path. It was getting stomped in some circumstances.
Fixes #38

PR #20, opened 2014-11-27
list directives in a list instead of a long string

givanse:
Same intent as in https://github.com/rwjblue/ember-cli-content-security-policy/pull/15, but coming from a branch and with a couple of tweaks.

Instead of:

    'script-src': "'self' https://cdn.mxpnl.com https://cdn.foobar.com"

Do:

    'script-src': ["'self'",
                   "https://cdn.mxpnl.com",
                   "https://cdn.foobar.com"]

Handles:

    contentSecurityPolicy: {
        'style-src': ["'self'", "'unsafe-inline'"],
        'frame-src': "'self' http://disqus.com"                                    
    }

And prints a deprecation warning to the server.

PR #54, opened 2015-12-12
A few improvements

### LINKED ISSUE #21: List policies in a list instead of a long string
Following the suggestion received on PR #15 I'm now opening this issue to track the development of this discussion.

@givanse [suggested](https://github.com/rwjblue/ember-cli-content-security-policy/pull/15#issue-46017797) using an array of strings to represent the single CSP rule — this allows to programmatically manipulate the various rules.

@rwjblue [liked the idea](https://github.com/rwjblue/ember-cli-content-security-policy/pull/15#issuecomment-59416397) and mentioned the fact that we'd probably better define a deprecation path.

[I added](https://github.com/rwjblue/ember-cli-content-security-policy/pull/15#issuecomment-61302427) that maybe we could use `csp:` schema-prefixed keywords so that there's no need to mix single quotes (`'`) and double quotes (`"`) and to allow, once more, programmability.

### LINKED ISSUE #38: "script-src 'self'" causes CSP violation with Ember inspector
Setting `contentSecurityPolicy: {'script-src': "'self'"}` on an Ember-CLI app causes a CSP violation, if that app is accessed on Windows (maybe also other OS) via latest stable Firefox with Ember Inspector addon installed. Here are the steps to reproduce:
* On Windows (7) generate a new Ember app (latest ember-cli) and start: `ember serve --host localhost`
* Add `contentSecurityPolicy: {'script-src': "'self'"}` to your `environment.js`
* On latest stable Firefox install addon [Ember Inspector] (https://addons.mozilla.org/en-US/firefox/addon/ember-inspector/) (make sure that its tab appears at the inspector pane)
* browse to `http://localhost:4200/`
* you will get the CSP report. If not, click the reload button

I know that adding `contentSecurityPolicy: {'script-src': "'self'"}` to `environment.js` doesn't really make sense, because that's the default setting anyway. But nevertheless I think that CSP violation shouldn't occur. By the way, it doesn't happen with Chrome + Ember Inspector. Here is the CSP report:
```json
{"csp-report":{
  "blocked-uri":"self",
  "document-uri":"http://localhost:4200/",
  "line-number":58,
  "original-policy":
    "script-src http://localhost:4200 http://localhost:35729 http://0.0.0.0:35729; 
     default-src 'none'; 
     font-src http://localhost:4200; 
     connect-src http://localhost:4200 ws://localhost:35729 ws://0.0.0.0:35729 http://localhost:4200/csp-report; 
     img-src http://localhost:4200; 
     style-src http://localhost:4200; 
     media-src http://localhost:4200; 
     report-uri http://localhost:4200/csp-report",
  "referrer":"",
  "script-sample":"call to eval() or related function blocked by CSP",
  "source-file":"resource://gre/modules/commonjs/toolkit/loader.js ->
                 resource://gre/modules/commonjs/sdk/loader/sandbox.js ->
                 resource://ember-inspector-at-emberjs-dot-com/ember-inspector/data/content-script.js",
  "violated-directive":"script-src http://localhost:4200 http://localhost:35729 http://0.0.0.0:35729"
}}
```
If I remove `contentSecurityPolicy: {'script-src': "'self'"}`, the CSP violation doesn't occur any longer.

sandstrom:
- Support serving CSP via meta tag.
- Break out policy generation into utility function.
- Enable overriding the default policy by setting a directive to `null`.
- Allow specifying the source list as an array.

None of these changes are breaking.

Fixes #21 and fixes #38.
Supersedes #51, #39, #41 and #20.

@rwjblue Let me know if you want me to change anything!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,17 +17,19 @@ used from your projects configuration:
 * `contentSecurityPolicy` -- This is an object that is used to build the final header value. Each key/value
   in this object is converted into a key/value pair in the resulting header value.
 
+* `contentSecurityPolicyMeta` -- Boolean. Toggle delivery via meta-tag. Useful for deployments where headers are not available (mobile, S3, etc) or to tether the CSP policy to the client payload (i.e. policy can be updated without reconfiguring servers). Check the W3C resource for details.
+
 The default `contentSecurityPolicy` value is:
 
 ```javascript
   contentSecurityPolicy: {
-    'default-src': "'none'",
-    'script-src': "'self'",
-    'font-src': "'self'",
-    'connect-src': "'self'",
-    'img-src': "'self'",
-    'style-src': "'self'",
-    'media-src': "'self'"
+    'default-src': ["'none'"],
+    'script-src':  ["'self'"],
+    'font-src':    ["'self'"],
+    'connect-src': ["'self'"],
+    'img-src':     ["'self'"],
+    'style-src':   ["'self'"],
+    'media-src':   ["'self'"]
   }
 ```
 
@@ -37,6 +39,8 @@ Which is translated into:
 default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self';
 ```
 
+If a directive is omitted it will default to `'self'`. To clear a directive from the default policy above, set it to `null`. The browser will fallback to the `default-src` if a directive does not exist.
+
 ### Example
 
 If your site uses **Google Fonts**, **Mixpanel**, a custom API at **custom-api.local** and a jQuery plugin which modifies the inline `style` attribute of some elements:
@@ -45,12 +49,12 @@ If your site uses **Google Fonts**, **Mixpanel**, a custom API at **custom-api.l
 // config/environment.js
 ENV.contentSecurityPolicy = {
   'default-src': "'none'",
-  'script-src': "'self' https://cdn.mxpnl.com", // Allow scripts from https://cdn.mxpnl.com
-  'font-src': "'self' http://fonts.gstatic.com", // Allow fonts to be loaded from http://fonts.gstatic.com
-  'connect-src': "'self' https://api.mixpanel.com http://custom-api.local", // Allow data (ajax/websocket) from api.mixpanel.com and custom-api.local
+  'script-src': ["'self'", "https://cdn.mxpnl.com"], // Allow scripts from https://cdn.mxpnl.com
+  'font-src': ["'self'", "http://fonts.gstatic.com"], // Allow fonts to be loaded from http://fonts.gstatic.com
+  'connect-src': ["'self'", "https://api.mixpanel.com", "http://custom-api.local"], // Allow data (ajax/websocket) from api.mixpanel.com and custom-api.local
   'img-src': "'self'",
-  'style-src': "'self' 'unsafe-inline' http://fonts.googleapis.com", // Allow inline styles and loaded CSS from http://fonts.googleapis.com 
-  'media-src': "'self'"
+  'style-src': ["'self'", "'unsafe-inline'", "http://fonts.googleapis.com"], // Allow inline styles and loaded CSS from http://fonts.googleapis.com
+  'media-src': null // `media-src` will be omitted from policy, browser will fallback to default-src for media resources.
 }
 ```
 
--- a/index.js
+++ b/index.js
@@ -1,27 +1,64 @@
 var chalk = require('chalk');
 
+var buildPolicyString = require('./lib/utils')['buildPolicyString'];
+
+var CSP_SELF        = "'self'";
+var CSP_NONE        = "'none'";
+var REPORT_PATH     = '/csp-report';
+
+var CSP_HEADER              = 'Content-Security-Policy';
+var CSP_HEADER_REPORT_ONLY  = 'Content-Security-Policy-Report-Only';
+
+var CSP_REPORT_URI          = 'report-uri';
+var CSP_FRAME_ANCESTORS     = 'frame-ancestors';
+var CSP_SANDBOX             = 'sandbox';
+
+var META_UNSUPPORTED_DIRECTIVES = [
+  CSP_REPORT_URI,
+  CSP_FRAME_ANCESTORS,
+  CSP_SANDBOX,
+];
+
+var unsupportedDirectives = function(policyObject) {
+  return META_UNSUPPORTED_DIRECTIVES.filter(function(name) {
+    return policyObject && (name in policyObject);
+  });
+}
+
+var appendSourceList = function(policyObject, name, sourceList) {
+  var oldSourceList;
+  var oldValue = policyObject[name];
+
+  if (Array.isArray(oldValue)) {
+    oldSourceList = oldValue;
+  } else if (!oldValue) {
+    oldSourceList = [];
+  } else if (typeof oldValue === 'string') {
+    oldSourceList = oldValue.split(' ');
+  } else {
+    throw new Error('Unknown source list value');
+  }
+
+  oldSourceList.push(sourceList);
+  policyObject[name] = oldSourceList.join(' ');
+}
+
 module.exports = {
   name: 'ember-cli-content-security-policy',
 
-  config: function(environment /*, appConfig */) {
-    var ENV = {
-      contentSecurityPolicyHeader: 'Content-Security-Policy-Report-Only',
+  config: function(/* environment, appConfig */) {
+    return {
+      contentSecurityPolicyHeader: CSP_HEADER_REPORT_ONLY,
       contentSecurityPolicy: {
-        'default-src': "'none'",
-        'script-src': "'self'",
-        'font-src': "'self'",
-        'connect-src': "'self'",
-        'img-src': "'self'",
-        'style-src': "'self'",
-        'media-src': "'self'"
+        'default-src':  CSP_NONE,
+        'script-src':   CSP_SELF,
+        'font-src':     CSP_SELF,
+        'connect-src':  CSP_SELF,
+        'img-src':      CSP_SELF,
+        'style-src':    CSP_SELF,
+        'media-src':    CSP_SELF
       }
     };
-
-    if (environment === 'development') {
-      ENV.contentSecurityPolicy['script-src'] = ENV.contentSecurityPolicy['script-src'] + " 'unsafe-eval'";
-    }
-
-    return ENV;
   },
 
   serverMiddleware: function(config) {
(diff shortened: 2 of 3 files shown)


sandstrom:
@rwjblue would you mind taking a look at this? :christmas_tree:

index.js
```diff
@@ -1,24 +1,50 @@
...
+        'connect-src':  CSP_SELF,
+        'img-src':      CSP_SELF,
+        'style-src':    CSP_SELF,
+        'media-src':    CSP_SELF
       }
     };
 
     if (environment === 'development') {
-      ENV.contentSecurityPolicy['script-src'] = ENV.contentSecurityPolicy['script-src'] + " 'unsafe-eval'";
+      ENV.contentSecurityPolicy['script-src'] += (' ' + CSP_UNSAFE_EVAL);
```
rwjblue: I don't think this is needed anymore. Ember CLI hasn't used evaled strings (for fake source maps) for a very long time...

sandstrom: @rwjblue Got it! I only kept it because it was already in the addon. Removed now.

index.js
```diff
@@ -70,11 +92,35 @@ module.exports = {
...
+      var policyHeader = appConfig.contentSecurityPolicyHeader;
+      var policyString = buildPolicyString(policyObject);
+
+      unsupportedDirectives(policyObject).forEach(function(name) {
+        var msg = 'CSP deliverd via meta does not support `' + name + '`, ' +
+                  'per the W3C recommendation.';
+        console.log(chalk.yellow(msg));
+      });
+
+      if (!policyHeader || policyHeader !== CSP_HEADER) {
```
rwjblue: I don't fully understand this conditional. I thought the meta scenario was for when you couldn't set headers? Why is the header required?

sandstrom: You are right, this is unnecessary. I'll fix it :smile:

rwjblue:
This seems good to me. Will try to merge and release tomorrow.

buschtoens:
Using the example you provided in the readme, I get very weird policy rules...

```js
  /**
   * ember-cli-content-security-policy
   */
  ENV.contentSecurityPolicyMeta = true;
  ENV.contentSecurityPolicy = {
    'default-src': "'none'",
    'script-src': ["'self'", "https://cdn.mxpnl.com"], // Allow scripts from https://cdn.mxpnl.com
    'font-src': ["'self'", "http://fonts.gstatic.com"], // Allow fonts to be loaded from http://fonts.gstatic.com
    'connect-src': ["'self'", "https://api.mixpanel.com", "http://custom-api.local"], // Allow data (ajax/websocket) from api.mixpanel.com and custom-api.local
    'img-src': "'self'",
    'style-src': ["'self'", "'unsafe-inline'", "http://fonts.googleapis.com"], // Allow inline styles and loaded CSS from http://fonts.googleapis.com
    'media-src': null // `media-src` will be omitted from policy, browser will fallback to default-src for media resources.
  }
```

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self' https://cdn.mxpnl.com e l f '; font-src 'self' http://fonts.gstatic.com e l f '; connect-src 'self' https://api.mixpanel.com http://custom-api.local l f '; img-src 'self'; style-src 'self' 'unsafe-inline' http://fonts.googleapis.com l f '; ">
``` 

```
[jan@arch community-app]$ ember v
version: 2.2.0-beta.1
Could not find watchman, falling back to NodeWatcher for file system events.
Visit http://www.ember-cli.com/user-guide/#watchman for more info.
node: 5.3.0
npm: 2.14.10
os: linux x64
```
If it would work, I'd be so much :+1: for this PR!

index.js
```diff
@@ -1,27 +1,46 @@
...
-        'img-src': "'self'",
-        'style-src': "'self'",
-        'media-src': "'self'"
+        'default-src':  CSP_NONE,
+        'script-src':   CSP_SELF,
+        'font-src':     CSP_SELF,
+        'connect-src':  CSP_SELF,
+        'img-src':      CSP_SELF,
+        'style-src':    CSP_SELF,
+        'media-src':    CSP_SELF
```
buschtoens: Found the bug. The lines should all be wrapped in an array like so:

```js
  config: function(/* environment, appConfig */) {
    return {
      contentSecurityPolicyHeader: CSP_HEADER_REPORT_ONLY,
      contentSecurityPolicy: {
        'default-src':  [CSP_NONE],
        'script-src':   [CSP_SELF],
        'font-src':     [CSP_SELF],
        'connect-src':  [CSP_SELF],
        'img-src':      [CSP_SELF],
        'style-src':    [CSP_SELF],
        'media-src':    [CSP_SELF]
      }
    };
  },
```

index.js
```diff
@@ -1,27 +1,46 @@
 var chalk = require('chalk');
 
+var buildPolicyString = require('./lib/utils')['buildPolicyString'];
+
+var CSP_SELF        = "'self'";
+var CSP_NONE        = "'none'";
```
buschtoens: Do not wrap them here though, as array are passed by reference and additional rules would add up.

buschtoens:
With this fix applied, everything seems to run perfectly fine. I really like the new semantics.
The only problem is, that backwards compatibility (i.e. everything in one long string) will be broken this way.

buschtoens:
Nope, another bug when using the `ENV.contentSecurityPolicyMeta = true` mode.

The livereload rules are only applied to the HTTP headers, not the meta tag. The are also re-calculated on *every* request. They definitely should be cached.

index.js
```diff
@@ -34,34 +53,35 @@ module.exports = {
...
       var headerConfig = appConfig.contentSecurityPolicy;
-      var normalizedHost = options.host === '0.0.0.0' ? 'localhost' : options.host;
 
       if (options.liveReload) {
         ['localhost', '0.0.0.0'].forEach(function(host) {
-          headerConfig['connect-src'] = headerConfig['connect-src'] + ' ws://' + host + ':' + options.liveReloadPort;
-          headerConfig['script-src'] = headerConfig['script-src'] + ' ' + host + ':' + options.liveReloadPort;
+          var liveReloadHost = host + ':' + options.liveReloadPort;
+          headerConfig['connect-src'] += (' ' + 'ws://' + liveReloadHost);
+          headerConfig['script-src'] += (' ' + liveReloadHost);
```
buschtoens: When wrapping `CSP_NONE` and `CSP_SELF` in arrays, these concatenations have to be replaced with `.push()`.

index.js
```diff
@@ -34,34 +53,35 @@ module.exports = {
...
         });
       }
 
-      if (header.indexOf('Report-Only')!==-1 && !('report-uri' in headerConfig)) {
-        headerConfig['connect-src'] = headerConfig['connect-src'] + ' http://' + normalizedHost + ':' + options.port + '/csp-report';
-        headerConfig['report-uri'] = 'http://' + normalizedHost + ':' + options.port + '/csp-report';
+      if (header.indexOf('Report-Only') !== -1 && !('report-uri' in headerConfig)) {
+        var normalizedHost = options.host === '0.0.0.0' ? 'localhost' : options.host;
+        var emberCliServer = 'http://' + normalizedHost + ':' + options.port + REPORT_PATH
+        headerConfig['connect-src'] += (' ' + emberCliServer);
```
buschtoens: Also has to be `.push()`.

On a related note. Should the `REPORT_PATH` really be part of `connect-src`? While Chrome doesn't throw any errors at me, it's not listed as [allowed syntax](http://content-security-policy.com/#source_list).

Furthermore, at least in the new ember-cli beta, `options.host` is `undefined`, when no `--host` option is passed to `ember serve`. Ironically, [that was me](https://github.com/ember-cli/ember-cli/pull/4781).

All in all, this should be:

```js
      if (header.indexOf('Report-Only') !== -1 && !('report-uri' in headerConfig)) {
        var normalizedHost = options.host || 'localhost';
        var emberCliServer = 'http://' + normalizedHost + ':' + options.port;
        headerConfig['connect-src'].push(emberCliServer);
        headerConfig['report-uri'] = emberCliServer + REPORT_PATH;
      }
```

rwjblue:
> The only problem is, that backwards compatibility (i.e. everything in one long string) will be broken this way.

We should be detecting a string and acting accordingly.  There should be no breaking change here.

@sandstrom - Would you mind looking at the above comments?

sandstrom:
@buschtoens thanks for pointing out these issues!

- The main culprit was the live-reload adjustment (its rule could not be appended if the existing setting was an array). This has been fixed. Good finding!
- Removed the report path, thanks for noting.
- The host normalization code has also been updated, thanks! (for breaking it with your patch :smile: and then pointing it out).
- Live-reload rules are still only applied to the HTTP headers. But I think this can be solved, will investigate.
- Regarding caching of live-reload rules, I don't think that's a performance issue. 

***

@rwjblue mostly done! Quick question: how is `options.liveReload` reached from within an addon's `contentFor` hook?

buschtoens:
@rwjblue 

> We should be detecting a string and acting accordingly. There should be no breaking change here.

I was gonna suggest the same, but it appears that `project.config(options.environment);` has problems with differing data types (in the new beta)? Or maybe I was just too decaffeinated to think straight.

---

@sandstrom Thanks for doing all that work. :smile: 

Regarding the caching: I did not mean to (only) cache the live-reload rules. I meant that it would be sensible to produce the CSP header string only once (during initialization) and store it thereafter. Currently `buildPolicyString(policyObject)` is executed with every single request. As far as I know, changes in the `config/environment.js` are ignored while `ember server` is running, but correct me if I'm wrong.

This way we could probably also include the live-reload rules in the meta tag. I'll have a glance add the addon hooks when I'm back home. :)

sandstrom:
@buschtoens If this is a performance drag I agree caching would be useful. I haven't benchmarked it, but an educated guess is that this isn't a performance bottleneck so no need to optimize. Processors are fast.

I've gotten the live-reload rules into the meta-tag btw.

README.md
```diff
@@ -17,17 +17,19 @@ used from your projects configuration:
 * `contentSecurityPolicy` -- This is an object that is used to build the final header value. Each key/value
   in this object is converted into a key/value pair in the resulting header value.
 
+* `contentSecurityPolicyMeta` -- Boolean. Toggle delivery via meta-tag. Useful for deployments where headers are not available (mobile, S3, etc) or to tether the CSP policy to the client payload (i.e. policy can be updated without reconfiguring servers). Check the W3C resource for details.
+
 The default `contentSecurityPolicy` value is:
 
 ```javascript
   contentSecurityPolicy: {
```
rwjblue: Should this section be updated to use an array also?

index.js
rwjblue: @sandstrom - Is this comment addressed?

sandstrom: Yes!

README.md
sandstrom: Done!

rwjblue:
Thanks for the hard work on this!  I'm going to land this into master and let it simmer for a short period (likely a few days).  Can y'all test out master and confirm things work as expected?

PR #55, opened 2015-12-30
[bugfix] use array syntax for default policy

sandstrom:
Solves an omission of mine in a previous PR. cc @rwjblue


rwjblue:
Thank you!

PR #57, opened 2016-01-01
Consider CSP's built-in fallback mechanism

sandstrom:
CSP has a built-in fallback mechanism. If, say, `connect-src` is not defined it will fall back to `default-src`. This can cause issues when the addon appends to an empty directive.

An example:

    Developer has has defined the following policy:
    `default-src: 'self' example.com;`
    
    An xhr to example.com (the connect-src directive) will be allowed. 
    Since no connect-src directive is defined it will fallback to default-src.
    
    Then an addon appends the connect-src entry live-reload.local, and the result is:
    `default-src: 'self' example.com; connect-src: live-reload.local;`
    
    After the addons change an xhr to example.com (which was previously permitted, 
    via fallback to default-src) is now rejected since it doesn't match 'live-reload.local'.

This PR fixes this by copying from default-src before appending to an empty entry.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+/* global require,module,process */
 var chalk = require('chalk');
 
 var buildPolicyString = require('./lib/utils')['buildPolicyString'];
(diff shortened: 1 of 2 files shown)


sandstrom:
@rwjblue eager to hear your thoughts on this!

It'll make the addon more pleasant to use. It avoids a scenario when CSP-adjustments (for live-reload, and in the future other addons) will cause unexpected behaviour.

rwjblue:
Thanks for explaining this in more detail over in slack.  This looks good to me.

PR #58, opened 2016-01-10
Restructure readme

sandstrom:
- Clarify that the meta-tag mode can be used for deployment to production
- Improve readability of examples
- Remove 'inline-script' and 'eval-script', they aren't a thing anymore
- Update install snippet
- Remove mention of 'wrapInEval' (no longer used in ember-cli)

@rwjblue As always, let me know if you have input on the proposed changes! :boat:


rwjblue:
Looks good to me!

PR #59, opened 2016-01-21
Fix: livereload violates csp if served via https

### LINKED ISSUE #43: csp-report not working on https

jelhan:
As mentioned in a comment to PR #44 also livereload violates csp if page is served via https. 9fd84a51f0e2b73a6eb405bba0a26167822ccff7 only fixed #43 (csp violation by csp report uri, which was actually funny) but not csp violation by livereload.

### DIFF
--- a/index.js
+++ b/index.js
@@ -102,7 +102,8 @@ module.exports = {
       if (options.liveReload) {
         ['localhost', '0.0.0.0'].forEach(function(host) {
           var liveReloadHost = host + ':' + options.liveReloadPort;
-          appendSourceList(policyObject, 'connect-src', 'ws://' + liveReloadHost);
+          var liveReloadProtocol = options.ssl ? 'wss://' : 'ws://';
+          appendSourceList(policyObject, 'connect-src', liveReloadProtocol + liveReloadHost);
           appendSourceList(policyObject, 'script-src', liveReloadHost);
         });
       }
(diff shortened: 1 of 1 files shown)


rwjblue:
@sandstrom - Mind reviewing?

bower.json
```diff
@@ -8,7 +8,7 @@
     "ember-qunit": "0.3.1",
     "ember-qunit-notifications": "0.0.7",
     "ember-resolver": "~0.1.15",
-    "jquery": "^1.11.1",
+    "jquery": "1.11.1",
```
sandstrom: Does this need to change? Why?

sandstrom:
@jelhan Good catch!

1. I don't think the jQuery dependency need to be changed.
2. You need to apply the same change at this line: 
  https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L151

*(eventually this concern will be merged into the ember-cli-inject-live-reload, see https://github.com/ember-cli/rfcs/pull/22)*

@rwjblue Looks good if my two points above are addressed. Though I don't have HTTPS on my local machine, so I cannot verify.

bower.json
jelhan: Have to update ember (1.13.12 / 2.2.2. / > 2.3.0) or stick to jquery < 2.x due to this bug http://emberjs.com/blog/2016/01/15/ember-2-3-released.html#toc_removing-the-jquery-version-assertion. But I don't want to address this in my PR. Just committed the change necessary for running tests by accident.

jelhan:
@sandstrom I don't think we have to change livereload port based on `options.ssl`. It's just about the protocol `ws:` vs `wss:`. Seems to work fine without any change to `liveReloadPort`.

sandstrom:
@jelhan Sorry if I was vague, I ment this line `appendSourceList(policyObject, 'connect-src', 'ws://' + liveReloadHost);` (there are two occurrences)

*(also, great if you can squash the commits together)*

bower.json
rwjblue: Yeah, the Ember assert is removed so we can update ember itself in a separate PR.

jelhan:
Fixed the missing one and squashed the commits together.

sandstrom:
@jelhan Great, thanks for finding and fixing this! :boat:

PR #60, opened 2016-02-04
Added the live reload host for setting of the security policies

webark:
incase you are using a custom host for live reload, or for the general host. I know this could be changing when the live reload addon gets completed, and it's pulling from the addon rather then here. but thought this would still be a nice stop gap incase people are using custom hosts.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,7 +100,7 @@ module.exports = {
       // can be moved to the ember-cli-live-reload addon if RFC-22 is implemented
       // https://github.com/ember-cli/rfcs/pull/22
       if (options.liveReload) {
-        ['localhost', '0.0.0.0'].forEach(function(host) {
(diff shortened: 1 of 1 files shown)


rwjblue:
Thanks!

sandstrom:
For this to work with the 'meta-mode' it would need a new ENV variable to be set by ember-cli-inject-live-reload. See https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/index.js#L152

@rwjblue is there a reason why only the port is available in inject-live-reload? i.e. would it be possible to add `EMBER_CLI_INJECT_LIVE_RELOAD_HOST`?

rwjblue:
@sandstrom - No reason other than that I probably just forgot :stuck_out_tongue_closed_eyes:

PR #68, opened 2016-09-14
Add `csp-headers` command

### LINKED ISSUE #11: Add command to output headers
It would be great if we could add a `csp-headers` command to `ember` to make copying the CSP-headers ridiculously easy. So I could just do:

```
ember csp-headers --environment production
```

And get the right headers for my production environment.

avdv:
Fixes #11.

### DIFF
--- a/index.js
+++ b/index.js
@@ -185,5 +185,9 @@ module.exports = {
         return '<meta http-equiv="' + CSP_HEADER + '" content="' + policyString + '">';
       }
     }
+  },
+
+  includedCommands: function() {
+    return require('./lib/commands');
   }
 };
--- a/README.md
+++ b/README.md
@@ -81,6 +81,22 @@ ENV.contentSecurityPolicy = {
 }
 ```
 
+## External Configuration
+
+In order to configure your production server, you can use the `csp-headers` command to obtain
+the current headers:
+
+```bash
+$ ember csp-headers --environment production --report-uri /csp-report
+
+# Content Security Policy Header Configuration
+#
+# for Apache: Header set Content-Security-Policy-Report-Only "..."
+# for Nginx : add_header Content-Security-Policy-Report-Only "...";
+
+default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; report-uri /csp-report;
+```
+
 *Please note*:
 + When running `ember serve` with live reload enabled, we also add the `liveReloadPort` to
   the `connect-src` and `script-src` whitelists.
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var buildPolicyString = require('./utils')['buildPolicyString'];
+var chalk = require('chalk');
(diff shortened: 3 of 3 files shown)


index.js
```diff
@@ -187,3 +187,52 @@ module.exports = {
...
+        ],
+
+        run: function(options) {
+          //this.ui.writeLine(JSON.stringify(options));
+          var path = require('path');
+          var fs = require('fs');
+          var merge = require('merge');
+          var env = options.environment;
+
+          // FIXME: how do I read the app config?
```
rwjblue: `this.project.config(options.environment)`

index.js
```diff
@@ -187,3 +187,52 @@ module.exports = {
...
+          var configPath = path.join(this.project.root, 'config/environment.js');
+
+          // FIXME: place the default config somewhere else so we can read it in the command?
```
rwjblue: Ya, that seems good

rwjblue:
Overall this looks good. I generally would prefer to move the commands to `lib/commands.js` and have the `includedCommands` hook look something like this:

```js
includedCommands: function() {
  return require('./commands');
}
```

Then have a `lib/commands.js` with the command added here.

avdv:
Thanks! I have implemented your changes, looks good already.

The only thing is, that the output is intermingled with some warnings and the "there's a new ember-cli version" message:

```
(node:31478) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
(node:31478) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
WARNING: WARNING: Node v6.5.0 has currently not been tested against Ember CLI and may result in unexpected behaviour.
version: 2.4.1

A new version of ember-cli is available (2.8.0).
default-src 'none'; script-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self'; media-src 'self'; report-uri /csp-report;
```
Is there a way to reliably disable any output except for the generated headers?

Otherwise I probably introduce an option to write into a given file...

rwjblue:
I would add some preamble prose a few new lines and the emit the actual headers (that makes it nice for human use), then start testing with the `--silent` flag and make sure it only emits the headers (which would make it nice for usage in shell pipelines).

avdv:
Alright, using `node ~/apps/node/bin/ember csp-headers --silent --environment production` produces the desired output.

avdv:
ping. Any objections? WDYT?

sandstrom:
@avdv Looks good to me! Nice work! I'll leave it to @rwjblue for a final decision though.

rwjblue:
I'll try to review again this week (sorry for delay)

avdv:
polite ping -- could you have a look again, thanks!

avdv:
@rwjblue /ping

(note that I have switched to an exponential backoff ping algorithm to avoid flooding your mailbox :wink:)

rwjblue:
Sorry about the super long delay here!

PR #70, opened 2017-01-20
Add the new middleware name from ember-cli

kratiahuja:
This addon should run before the files are served and after the build is done.


rwjblue:
My kingdom for comments in `package.json` 😿

rwjblue:
Thank you for updating!

PR #71, opened 2017-04-03
Don't include dist in published builds (2.5MB)

runspired:
Noticed while grepping for addons using `lookupFactory` that this addon was including it's `dist/` folder in published builds. I added the `files: []` array here with a rule to exclude `dist`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,9 @@
     "doc": "doc",
     "test": "tests"
   },
+  "files": [
+    "!dist/"
+  ],
   "scripts": {
     "start": "ember server",
     "build": "ember build",


rwjblue:
I have not tried adding a `files` array with only an exclusion, does this still include all other code (e.g. `lib/`)?

runspired:
@rwjblue I was actually wondering that and figured you might know and if so you'd point out my mistake ;)

I'll do some research.

runspired:
This suggests this bad: https://github.com/npm/npm/wiki/Files-and-Ignores#details

Will open a PR for .npmignore

PR #72, opened 2017-04-03
Don't include dist in published builds (2.5MB)

runspired:
Noticed while grepping for addons using `lookupFactory` that this addon was including it's `dist/` folder in published builds. I added the `dist/` rule here to `.npmignore` to fix this.  Replaces #71


rwjblue:
released as 0.6.1

PR #73, opened 2017-05-01
Add worker-src as a supported policy

Boubalou:
This will prevent getting CSP Error reports when using `ember-service-worker` such as this:

```[Report Only] Refused to create a worker from 'http://localhost:4200/sw.js' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'worker-src' was not explicitly set, so 'default-src' is used as a fallback.```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ The default `contentSecurityPolicy` value is:
     'connect-src': ["'self'"],
     'img-src':     ["'self'"],
     'style-src':   ["'self'"],
-    'media-src':   ["'self'"]
+    'media-src':   ["'self'"],
+    'worker-src':  ["'self'"]
   }
 ```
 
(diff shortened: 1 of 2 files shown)


sandstrom:
@Boubalou Can you describe the underlying issue you are trying to solve?

My read on this is that you're trying to patch a CSP-issue unique to your project by changing the addon defaults. But the defaults aren't ment to cover everything, it's just some sane defaults.

Please let me know if this is a misunderstanding on my part! 😄

Boubalou:
I believe you might be right! By enabling early the service worker process, it might have ignored my subsequent changes to the CSP config, which led me to think that `worker-src` had to be defined by default in this PR. 

Let me re-test this, and I will close if it was my mistake!

Boubalou:
Good catch @sandstrom !

Issue #76, opened 2017-08-02
Example for development options


pcambra:
Maybe it would be useful to include in the README a open to all development example, something like:

```
if (environment === 'development') {
    ENV.contentSecurityPolicy = {
      'default-src': ["*"],
      'script-src':  null,
      'font-src':    null,
      'connect-src': null,
      'img-src':     null,
      'style-src':   null,
      'media-src':   null
    };
}
```

sandstrom:
@pcambra Can you PR something for the *Examples* section of the readme?

jelhan:
I would recommend disabling or uninstalling it all. I don't see the use case for keeping it enabled but allowing all requests. Maybe this one was just motivated by the confusing configuration options? Could this be closed as there is an `enabled` option (#94) now?

rwjblue:
Ya, agreed @jelhan

PR #77, opened 2017-08-07
Added development configuration example

pcambra:
As per #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,24 @@ ENV.contentSecurityPolicy = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


sandstrom:
@pcambra Thanks!

I've added a commit with some suggested changes, let me know what you think!

jelhan:
This is outdated by #94. It could be disabled now by setting `ENV['ember-cli-content-security-policy'].enabled = false` which is much more inline with the rest of the ecosystem.

PR #78, opened 2017-08-14
Update to ember-cli 2.14 blueprint.

rwjblue:
This will ultimately require a 0.7.0 (since we are officially dropping support for Node < 4).


package.json
```diff
@@ -2,44 +2,49 @@
...
-  },
-  "repository": "https://github.com/rwjblue/ember-cli-content-security-policy",
-  "engines": {
-    "node": ">= 0.10.0"
-  },
-  "author": "",
-  "license": "MIT",
+    "start": "ember server",
+    "test": "ember try:each"
+  }
```
rwjblue: great catch, thank you!

rwjblue:
Thank you for reviewing (and fixing)!

PR #83, opened 2018-10-31
Generate package-lock.json

YoranBrondsema:
To fix the versions of all dependencies.


rwjblue:
I'd prefer to use `yarn` instead, mind swapping to that?

YoranBrondsema:
Yup! See #85. Out of curiosity, do you have any reasons, technical or not, to prefer yarn over npm?

PR #84, opened 2018-10-31
Add option to output raw csp (fix #81)

### LINKED ISSUE #81: Difficult to extract a usable CSP using csp-headers command
We are working on automating our CSP generation during deployment and ran into a couple of small usability issues that I wanted to report to see what your take is on this kind of thing.

1. Since the `ember csp-headers` command writes the CSP to `stderr`, we end up having to strip out any potential deprecation warnings from the output in order to get a valid CSP extracted:

*Sample output:*
```
$ ember csp-headers
DEPRECATION: Overriding init without calling this._super is deprecated. Please call `this._super.init && this._super.init.apply(this, arguments);` addon: `ember-data`
    at Function.Addon.lookup (/Users/jasonr/FrontEnd/node_modules/ember-cli/lib/models/addon.js:1617:27)
# Content Security Policy Header Configuration
#
# for Apache: Header set Content-Security-Policy "..."
# for Nginx : add_header Content-Security-Policy "...";

default-src self;
```

2. Additionally, even if the output was to be moved to `stdout` instead of `stderr`, we would still need to strip out the preceding instructional content:

```
# Content Security Policy Header Configuration
#
# for Apache: Header set Content-Security-Policy "..."
# for Nginx : add_header Content-Security-Policy "...";
```

### Proposal:

1. Ensure the resulting CSP content is written to `stdout` instead of `stderr`
2. Add a flag to suppress the instructional content (or move the instructional content to `--help` instead of always outputting it during csp generation)

Perhaps something like this:

```
$ ember csp-headers --csp-only
default-src self;
```

This way the output is easily pipeable into other scripts, readable from childProcess.exec, and generally just ready to be used as soon as it is returned to the caller.

Thanks for looking :) I can probably take a stab at it sometime soon if you agree that a change like this is desirable.

YoranBrondsema:
1. Fixes #81. With a `--raw` option, the configuration guidelines are omitted.
2. Output to the standard output rather than to error. Maybe there's a good reason for sending it to error. In that case this change can be reverted. But otherwise, it seems semantically wrong to do that.
3. Add a small description for the `report-uri` option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "4"
+  - '8'
 
 sudo: false
 
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -19,8 +19,15 @@ module.exports = {
       },
       {
         name: 'report-uri',
-        type: String
-      }
+        type: String,
+        description: 'Sets report-uri for the policy'
+      },
+      {
+        name: 'silent',
+        type: Boolean,
+        default: false,
+        description: 'Only outputs the policy without the instructions for Apache and Nginx'
+      },
     ],
 
     run: function(options) {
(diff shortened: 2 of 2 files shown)


jelhan:
Did you considered `--silent` flag? `ember test` is already supporting that one.

In contrast to `--quiet` flag `--silent` is often used as the opposite of `--verbose`: do only print minimal data user has asked for. An example is `curl`:
>        -s, --silent
>              Silent  or  quiet  mode.  Don't  show progress meter or error messages.
>              Makes Curl mute. It will still output the data you ask for, potentially even
>              to the terminal/stdout unless you redirect it.

YoranBrondsema:
Thanks for your suggestion @jelhan. The semantics of `--silent` indeed correspond better.

@rwjblue: is it possible to get your feedback so that we can get this PR merged?

sandstrom:
@YoranBrondsema Looks good!

However, I'm hesitant to merge due to the stderr -> stdout switch. While I think the change makes sense, maybe there is a reason for it that I don't know about.

So I'd like to check with @rwjblue to see that there aren't a special reason for outputting to stderr.

rwjblue:
@sandstrom is correct, emitting to stdout breaks things that attempt to parse the console output for test pass/fail. However, it may not be a factor in our case (e.g. if we only emit during development but not tests), right?

YoranBrondsema:
@rwjblue OK I understand the reasoning behind sending to stderr. The motivation to send to stdout is that

1. I was surprised that I couldn't read the output when consuming by a Ruby script. Only when I dug into the code did I see that it was sending to stderr, so I had to adapt the script accordingly.
2. It doesn't "feel" right as we're not outputting some error message, we're just displaying some configuration information. So semantically I think stdout makes more sense than stderr.

Either way, I realize now that this change does not belong in this PR. So I will undo it, so that this PR only adds the `--silent` flag.

sandstrom:
@YoranBrondsema I spoke with rwjblue about this. I interpreted his reasoning like this:

For things that are consumed by tests, we cannot use stdout since it will mess with the pass/fail check. However, since this command won't be used during tests, it's fine to switch.

So, in other words, feel free to open a new PR with the switch from stderr -> stdout. If you can, write a line or two about whether that would reasonably affect any tests and ping rwjblue in that PR.

sandstrom:
@YoranBrondsema I don't know why one travis job fails, but I'm willing to merge anyway because I cannot see that anything changed in this PR could cause those errors.

Thanks for your work on this! 🏅

YoranBrondsema:
I tried re-running it but it seems to consistently fail on the beta release of Ember.js. Thanks for the merge!

PR #85, opened 2018-10-31
Add yarn.lock

YoranBrondsema:
Following comment on #83.


rwjblue:
😭 - Looks like we have to do some general updating here, the failures in CI are due to using Phantom and testing on versions of Ember that require things _like_ `WeakMap`...

sandstrom:
Fixed by #87

PR #86, opened 2018-11-11
Use Chrome in tests instead of PhantomJS

YoranBrondsema:


.travis.yml
```diff
@@ -3,40 +3,67 @@ language: node_js
...
   # so that your addon works for all apps
-  - "4"
+  - "6"
 
 sudo: false
+dist: trusty
 
+addons:
+  chrome: stable
+<% if (yarn) { %>
```
rwjblue: I think these conditionals are causing problems with the parsing in CI. I think you should drop them in favor of the `yarn` side of each...

sandstrom:
I think this was fixed by #87

PR #88, opened 2019-02-15
whitelist script tag inserted for test file loaded assertion using nonce

jelhan:
This adds a nonce to <script> tag inserted by ember-cli for assertion that test file was loaded. Otherwise this violates default CSP in tests.

Nonce is always present in CSP header but only part of CSP meta tag injected into `tests/index.html`.

A nonce used in productive applications must be cryptographically random. But since this one is only used in local development and if executing tests, using a static string is fine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,6 +100,8 @@ module.exports = {
     var options = config.options;
     var project = options.project;
 
+    console.log(options.environment);
+
     app.use(function(req, res, next) {
       var appConfig = project.config(options.environment);
 
@@ -111,6 +113,8 @@ module.exports = {
         return;
       }
 
+      // the local server will never run for production builds, so no danger in adding the nonce all the time
+      // even so it's only needed if tests are executed by opening `http://localhost:4200/tests`
       if (policyObject) {
         appendSourceList(policyObject, 'script-src', "'nonce-" + STATIC_TEST_NONCE + "'");
       }
--- a/index.js
+++ b/index.js
@@ -23,6 +23,8 @@ var META_UNSUPPORTED_DIRECTIVES = [
   CSP_SANDBOX,
 ];
 
+var NONCE = 'abcdefg';
+
 var unsupportedDirectives = function(policyObject) {
   return META_UNSUPPORTED_DIRECTIVES.filter(function(name) {
     return policyObject && (name in policyObject);
@@ -109,6 +111,10 @@ module.exports = {
         return;
       }
 
+      if (policyObject) {
+        appendSourceList(policyObject, 'script-src', "'nonce-" + NONCE + "'");
+      }
+
       // can be moved to the ember-cli-live-reload addon if RFC-22 is implemented
       // https://github.com/ember-cli/rfcs/pull/22
       if (options.liveReload) {
@@ -159,7 +165,7 @@ module.exports = {
     });
   },
 
-  contentFor: function(type, appConfig) {
+  contentFor: function(type, appConfig, existingContent) {
     if ((type === 'head' && appConfig.contentSecurityPolicyMeta)) {
       var policyObject = appConfig.contentSecurityPolicy;
       var liveReloadPort = process.env.EMBER_CLI_INJECT_LIVE_RELOAD_PORT;
@@ -173,6 +179,10 @@ module.exports = {
           appendSourceList(policyObject, 'connect-src', 'wss://' + liveReloadHost);
           appendSourceList(policyObject, 'script-src', liveReloadHost);
         });
+      }
+
+      if (policyObject && appConfig.environment === 'test') {
+        appendSourceList(policyObject, 'script-src', "'nonce-" + NONCE + "'");
       }
 
       var policyString = buildPolicyString(policyObject);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -23,6 +23,8 @@ var META_UNSUPPORTED_DIRECTIVES = [
   CSP_SANDBOX,
 ];
 
+var NONCE = 'abcdefg';
```
sandstrom: How about `STATIC_TEST_NONCE`?

index.js
```diff
@@ -109,6 +111,10 @@ module.exports = {
         return;
       }
 
+      if (policyObject) {
```
sandstrom: For clarity, how about checking for `options.environment === 'test'` here too

jelhan:
> I haven't used this addon in tests much, so I cannot really vouch for whether this will work or not.

Steps to check this feature:

- `ember new ember-test-csp && cd ember-test-csp`
- `ember install ember-cli-content-security-policy`
- `ember serve` and open `http://localhost:4200/tests` in browser

You will notice two CSP violations reported:

```
Content Security Policy violation:
{
  "csp-report": {
    "document-uri": "http://localhost:4200/tests?hidepassed",
    "referrer": "",
    "violated-directive": "script-src-elem",
    "effective-directive": "script-src-elem",
    "original-policy": "default-src 'none'; script-src 'self' localhost:4200 0.0.0.0:4200 undefined:4200; font-src 'self'; connect-src 'self' ws://localhost:4200 ws://0.0.0.0:4200 ws://undefined:4200 http://localhost:4200; img-src 'self'; style-src 'self'; media-src 'self'; report-uri http://localhost:4200/csp-report;",
    "disposition": "report",
    "blocked-uri": "inline",
    "line-number": 38,
    "source-file": "http://localhost:4200/tests?hidepassed",
    "status-code": 200,
    "script-sample": ""
  }
}

Content Security Policy violation:
{
  "csp-report": {
    "document-uri": "http://localhost:4200/tests?hidepassed",
    "referrer": "",
    "violated-directive": "style-src-attr",
    "effective-directive": "style-src-attr",
    "original-policy": "default-src 'none'; script-src 'self' localhost:4200 0.0.0.0:4200 undefined:4200; font-src 'self'; connect-src 'self' ws://localhost:4200 ws://0.0.0.0:4200 ws://undefined:4200 http://localhost:4200; img-src 'self'; style-src 'self'; media-src 'self'; report-uri http://localhost:4200/csp-report;",
    "disposition": "report",
    "blocked-uri": "inline",
    "line-number": 6012,
    "column-number": 23,
    "source-file": "http://localhost:4200/assets/test-support.js",
    "status-code": 200,
    "script-sample": ""
  }
}
```

The first one should be fixed by this PR, the second one is addressed by https://github.com/qunitjs/qunit/pull/1369.

index.js
jelhan: Actually this needs to be applied also for `options.environment === 'development'` cause tests may be executed by `http://localhost:4200/tests`. In that case `option.environment` is true. So would end up checking the request target also...

sandstrom: @jelhan Alright, I haven't used `http://localhost:4200/tests` and I believe you, but it sounds weird that they aren't running under the test environment. Is that by design?

Anyway, perhaps we could test for both then? 

```js
options.environment === 'test' && options.environment === 'development'

```

jelhan: It's a complex topic for `http://localhost:4200/tests` and server middleware. The tests are running under test environment but the middleware is running under the environment used to start the server. So if we want to make sure that the nonce is only present in headers for testing we would end up with a check like this:

```js
options.environment === 'test' || req.path === '/tests'
```

To be honest I haven't checked if `ember test --server` runs middleware with `test` environment but I expect it to do so.

But I don't see a big issue in having that nonce present always in request headers. They don't influence build result but only used for development server (`ember serve`).

jelhan: Have done some quick tests and it seems like `serverMiddleware` hook is not executed if running `ember test --server`. CSP headers are not present at all.

sandstrom: @jelhan Got it, makes sense! I know the test server is only running under development and won't affect builds. I'm just thinking that making that clear in the if-statement will make it easier to reason about the code.

But you could also add a comment instead, something like:

```js
// the local server will never run for production builds, so no danger in adding the nonce all the time
```

*(I still would have gone with an if-statement, but I don't have a strong opinion on this)*

sandstrom:
@jelhan Seems like some of the tests failed, would you like to have a look at it? https://travis-ci.com/rwjblue/ember-cli-content-security-policy/jobs/178664703

You can try to just force-push to re-trigger another test-run, sometimes it's because of timeouts.

@rwjblue I don't have any further comments on this PR. In my opinion it's good to merge. But happy to hear your thoughts before I do anything.

jelhan:
> @jelhan Seems like some of the tests failed, would you like to have a look at it? https://travis-ci.com/rwjblue/ember-cli-content-security-policy/jobs/178664703

I don't think that issue for `ember-beta` and `ember-canary` is related to my changes. Seeing the same issues in #89. Actually it doesn't seem like there are any tests except for linting.

index.js
jelhan: :+1:

index.js
```diff
@@ -98,6 +100,8 @@ module.exports = {
     var options = config.options;
     var project = options.project;
 
+    console.log(options.environment);
```
rwjblue: Whooops, I missed this before I merged!

jelhan: Arghs and I missed that I committed that one with the last inline comment. I'm very sorry about that. Do you want me to open a PR to remove it or are you doing it yourself?

jelhan: `index.js` is not covered by ESLint :sob:

sandstrom: I've removed the `console.log` call in https://github.com/rwjblue/ember-cli-content-security-policy/commit/97cfed04ea213b9f1ac2c2f18a5c3efec67a730f (hope that was okay @rwjblue)

jelhan: @sandstrom Thanks a lot. I'm really sorry for the churn.

sandstrom: @jelhan No worries, thanks for making this code better! 😄

PR #89, opened 2019-02-16
fix: undefined:49153 is whitelisted in CSP, if `options.liveReloadHost` is not set

jelhan:


sandstrom:
@rwjblue Looks good to me! 

@jelhan There seems to be an issue with travis though

Issue #92, opened 2019-02-25
refactor configuration


jelhan:
The configuration of this addon doesn't feel inline with the rest of the ecosystem. This gets even worse if new configuration option are needed as it's the case for #91.

I like to suggest to refactor the configuration API as the following:

```js
ENV['ember-cli-content-security-policy'] = {
  enabled: true,
  metaTag: false,
  policies: {
    'default-src': ["'none'"],
    'script-src':  ["'self'"],
    'font-src':    ["'self'"],
    'connect-src': ["'self'"],
    'img-src':     ["'self'"],
    'style-src':   ["'self'"],
    'media-src':   ["'self'"]
  },
  reportOnly: true,
}
```

The values shown above are suggested defaults.

For a transition period we might want to support existing configuration API. In that case the defaults would look like:

```js
ENV['ember-cli-content-security-policy'] = {
  enabled: true,
  metaTag: ENV.contentSecurityPoliyMeta,
  policies: ENV.contentSecurityPolicy || {
    'default-src': ["'none'"],
    'script-src':  ["'self'"],
    'font-src':    ["'self'"],
    'connect-src': ["'self'"],
    'img-src':     ["'self'"],
    'style-src':   ["'self'"],
    'media-src':   ["'self'"]
  },
  reportOnly: ENV.contentSecurityPolicyHeader === 'Content-Security-Policy-Report-Only',
}
```

If we consider disabling the addon by setting `ENV.contentSecurityPolicy === {}` as a public API as suggested in #77, we might want to change the default for `enabled` to `!ENV.contentSecurityPolicy || Object.keys(ENV.contentSecurityPolicy).length !== 0`. We should deprecate the old configuration API at the same time. In that case this wouldn't be a breaking change but add some additional maintenance costs.

Please note that this API wouldn't require a change for #91. It would allow to disable throwing in tests by setting `enabled = environment !== 'test'`. `metaTag` option would be forced to `true` if running tests.

jelhan:
Just recognized that this isn't runtime configuration but for build and server only. Due to this it should also be moved from `config/environment.js` to `ember-cli-build.js`.

lifeart:
is it possible to have more `dummers-frendly` and declarative configuration, like

```
    csp
        policies:
        
            images:
                local: true,
                raw-inline: true,
                fromHosts: [ 'http://...', 'http://...' ]
            fonts:
                local: true,
            styles:
                local: true,
                inline: true
                ...
```
[...]

Issue #16, opened 2014-10-25
tests (i know its a pain)


stefanpenner:

jelhan:
While the test coverage is still not great, we have some tests for basic functionality. So we could close this issue in my opinion.

PR #94, opened 2019-06-08
refactor configuration

jelhan:
Refactors configuration as discussed in #92. It aims to be backward-compatible. Also adds some tests and therefore partly addresses #16.

This is still a work-in-progress as I like to extend the test coverage. This addon is used a lot and was part of the default blueprint some time ago. Therefore I like to reduce the risk of introducing bugs as much as possible. Especially cause I don't want to harm it's reputation.

But I would love to get some early feedback. Haven't used `ember-cli-addon-tests` before and I'm not that familiar with ember-cli addon hooks.

This is preparation work needed for the additional features discussed in #91 and https://github.com/jelhan/ember-style-modifier/issues/11 in my opinion. For both we need to alter existing configuration and it seems to be very hard to do that with the configuration API as it is right now.


README.md
```diff
@@ -23,70 +23,68 @@ Installation
...
-
-```
-default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self';
-```
-
-If a directive is omitted it will default to `'self'`. To clear a directive from the default policy above, set it to `null`. The browser will fallback to the `default-src` if a directive does not exist.
+  ```
+  To clear a directive from the default policy, set it to `null`.
+  The browser will fallback to the `default-src` if a directive does not exist.
+- `reportOnly?: boolean`
```
sandstrom:
```suggestion
- `reportOnly: boolean`
```

README.md
```diff
@@ -23,70 +23,68 @@ Installation
...
+
+This addon is configured via your applications `config/environment.js` file using `ember-cli-content-security-policy` key:
+
+- `delivery?: string[]`
+  CSP is delivered via HTTP Header if includes `"header"` and via meta element if includes `"meta"`.
+  Defaults to `["header"]`.
+- `enabled?: boolean`
+  Controls if addon is enabled at all.
+  Defaults to `true`.
+- `policy?: object`
```
sandstrom:
```suggestion
- `policy: object`
```

README.md
```diff
@@ -23,70 +23,68 @@ Installation
...
-    'style-src':   ["'self'"],
-    'media-src':   ["'self'"]
+## Configuration
+
+This addon is configured via your applications `config/environment.js` file using `ember-cli-content-security-policy` key:
+
+- `delivery?: string[]`
+  CSP is delivered via HTTP Header if includes `"header"` and via meta element if includes `"meta"`.
+  Defaults to `["header"]`.
+- `enabled?: boolean`
```
sandstrom:
```suggestion
- `enabled: boolean`
```

README.md
```diff
@@ -23,70 +23,68 @@ Installation
...
-    'font-src':    ["'self'"],
-    'connect-src': ["'self'"],
-    'img-src':     ["'self'"],
-    'style-src':   ["'self'"],
-    'media-src':   ["'self'"]
+## Configuration
+
+This addon is configured via your applications `config/environment.js` file using `ember-cli-content-security-policy` key:
+
+- `delivery?: string[]`
```
sandstrom:
```suggestion
- `delivery: string[]`
```

README.md
```diff
@@ -23,70 +23,68 @@ Installation
...
+  CSP is delivered via HTTP Header if includes `"header"` and via meta element if includes `"meta"`.
```
sandstrom:
```suggestion
  CSP is delivered via HTTP Header if delivery includes `"header"` and via meta element if it includes `"meta"`.
```

index.js
```diff
@@ -100,14 +119,14 @@ module.exports = {
     app.use(function(req, res, next) {
       var appConfig = project.config(options.environment);
 
-      var header = appConfig.contentSecurityPolicyHeader;
-      var policyObject = appConfig.contentSecurityPolicy;
-
-      if (!header || !policyObject) {
+      if (!appConfig['ember-cli-content-security-policy'].enabled) {
```
sandstrom: This string occurs a few times, could move it into a constant, e.g. `const CONFIG_KEY = 'ember-cli-content-security-policy';`

node-tests/acceptance/deliver-test.js
```diff
@@ -0,0 +1,127 @@
+const expect = require('chai').expect;
+const denodeify = require('denodeify');
+const request = denodeify(require('request'));
+const AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
+const fs = require('fs-extra');
+
+const cspMetaTagRegExp = /<meta http-equiv="Content-Security-Policy" content=".*">/i;
```
sandstrom: Extremely minor (free free to keep it as is), but I think the JS convention is to use UPPER_CASE here:

```js
CSP_META_TAG_REG_EXP = /<meta http-equiv="Content-Security-Policy" content=".*">/i;
```

index.js
```diff
@@ -68,28 +68,47 @@ var appendSourceList = function(policyObject, name, sourceList) {
...
+      'to configure ember-cli-content-security-policy is deprecate and will be removed in v2.0.0. ' +
+      'Please find detailed information about new configuration options in addon documentation.',
+      !appConfig.contentSecurityPolicy || !appConfig.contentSecurityPolicyHeader || !appConfig.contentSecurityPolicyMeta
+    );
+
+    if (appConfig.contentSecurityPolicy) {
+      defaultConfig.policy = appConfig.contentSecurityPolicy;
+    }
+    if (appConfig.contentSecurityPolicyMeta === false) {
+      defaultConfig.delivery = ['header'];
```
sandstrom: Could constantize these values, e.g. `const DELIVERY_HEADER = 'header';`

jelhan:
@sandstrom Thanks for your quick review. Have addressed the feedback.

jelhan:
The PR does not yet address the refactoring of policy object as suggested by @lifeart in https://github.com/rwjblue/ember-cli-content-security-policy/issues/92#issuecomment-471951794. I think this should be left for follow up PR to limit the scope here.

sandstrom:
@jelhan Looks good! 😄 

Sorry if I was unclear regarding const/let/var. I've added a new commit with some changes:
https://github.com/rwjblue/ember-cli-content-security-policy/pull/94/commits/333f882b91d168adeb274a70bbde63f1c2080439

Feel free to squash everything together!

Regarding the tests, I haven't been able to inspect them really, but I'm happy to trust you that they'll work.

One final thing is that you need to remove `wip` from the title for the WIP check to pass. Other than that + squashing I think this is ready to go!

jelhan:
> One final thing is that you need to remove `wip` from the title for the WIP check to pass. Other than that + squashing I think this is ready to go!

Thanks for your trust. I would like to add some more tests to ensure we won't break existing apps. Will remove _WIP_ as soon as I'm feeling comfortable with test coverage.

sandstrom:
@jelhan (1) sounds good and (2) probably, but I don't know. 

In the broader JS community I know there are differences, some like to use `const` for every variable that shouldn't be mutated and others reserve it for constant-like values [more similar to how constants are used in other languages]. I'm leaning towards the latter, but others may think differently.

jelhan:
@sandstrom I think this one is ready to be merged. Still missing a lot of tests but while working on them I encountered other bugs/issues (https://github.com/jelhan/ember-cli-content-security-policy/pull/1) which should be addressed in separate PRs.

sandstrom:
@jelhan Awesome!

Would you like to tidy up the commits somewhat, before we merge?

For example squash `let/const` into `refactor var -> const/let` (so that they are 1 single commit). Similar with the edits to README and possibly some other commits as well. I don't mind multiple commits, that's fine! But each commit should be (mostly) coherent and isolated.

jelhan:
@sandstrom Thought you are following a squash + merge strategy. Mind if I squash everything into 1 single commit?

jelhan:
Squashed the commits a little bit along of the commit history. Would like to avoid reorganizing the changes for a nice commit history. So it's still 7 commits without well separation of concerns. Feel free to squash them all up into one while merging.

sandstrom:
@jelhan Maybe this project does [follow such a strategy], I don't really know. But will use squash + merge now.

Thanks for your hard work on this, much appreciated! 🙏

PR #95, opened 2019-06-10
remove trailing whitespace from csp string

jelhan:
CSP string generated and used in HTTP header, `<meta>` element and `ember csp-headers` command had a trailing whitespace. It's ugly and makes processing output of `ember csp-headers` more difficult in some cases.


sandstrom:
thanks @jelhan 🙏

PR #96, opened 2019-06-13
fix inconsistency between meta element and HTTP header regarding live reload support

jelhan:
Also adds a test that same CSP is delivered using HTTP header and meta element.

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,10 +75,10 @@ let appendSourceList = function(policyObject, name, sourceList) {
 let liveReloadOptions;
 
 // appends directives needed for Ember CLI live reload feature to policy object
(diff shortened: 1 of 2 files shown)


sandstrom:
@jelhan Looks good! I've made a minor change (check if it's enabled outside the function call, for clarity). Let me know what you think

jelhan:
Fixed `options` is `undefined` caused by refactoring. Ready to be merged from my side. Do you want me to squash or are you using the squash + merge button?

sandstrom:
@jelhan Good catch, thanks! 🎣 

I'll squash + merge.

PR #97, opened 2019-06-14
move configuration to ember cli build

### LINKED ISSUE #30: options should not come from config/env, rather the brocfil. These are build-server concerns not run-time concerns

jelhan:
Closes #30


jelhan:
I'm not 100%ly convinced by the unit tests for `calculateConfig()` as they are to tightly coupled with Implementation in my opinion. But on the other side the end-to-end tests using ember-cli-addon-tests are terrible slow... Would love to get some input on that one.

index.js
```diff
@@ -254,5 +200,76 @@ module.exports = {
...
+    this.calculateConfig(app);
+  },
+
+  // Configuration is only available by public API in `app` passed to `included` hook.
+  // We calculate configuration in `included` hook and use it in `serverMiddleware`
+  // and `contentFor` hooks, which are executed later. This is necessary cause Ember CLI
+  // does not provide a public API to read build time configuation (`ember-cli-build.js`)
+  // yet. `this._findHost(this).options` seems to be the only reliable way to get it in
+  // these hooks but is private API.
+  calculateConfig(app) {
```
rwjblue: Can we make this a "pure function", that we could then import from the test and call directly? It seems like we could test a ton of permutations that way, and keep the test coupling to implementation a bit smaller...

```js
// ...snip...
included(app) {
  this._config = calculateConfig(app.env, app.options || {}, this.project.config(app.env));
},

// ...snip...

// below main `module.exports = ` line

function calculateConfig(environment, buildConfig, runConfig, ui) {
  // current implementation
  return config;
}
module.exports._calculateConfig = calculateConfig;
```

Then if you do this, you don't need an "app mock", you can instead pass the various permutations of POJO's directly into the `calculateConfig` method.

jelhan:
@rwjblue Thanks for your feedback. Refactored as requested.

jelhan:
Noticed that policy object wasn't merged anymore. Fixed that one in last commit and added tests.

I'm not a fan of this behavior to be honest but removing by a bug isn't a solution I guess. :rofl:

index.js
```diff
@@ -66,7 +66,8 @@ let appendSourceList = function(policyObject, name, sourceList) {
     oldSourceList = oldValue;
   }
 
-  oldSourceList.push(sourceList);
+  // do not mutate existing source list to prevent leaking state
+  oldSourceList.slice().push(sourceList);
```
sandstrom: This doesn't make any sense to me. You are creating a copy of the array, pushing something into the copy and then throwing the copy away (all on the same line).

Am I missing something obvious? Or is this code a no-op?

jelhan: Good catch! Demonstrates how bad our test coverage still is.

This method is awful. Will add tests covering that one and refactor it in another PR.

jelhan:
@sandstrom This is ready to be merged from my side.

sandstrom:
@rwjblue Do you have any more comments on this?

rwjblue:
LGTM, thanks for working through it @jelhan!

Issue #100, opened 2019-06-30
Roadmap for v2.0


jelhan:
_Disclaimer: This is only a proposal as I'm not the maintainer of this repo._

As you might have noticed I'm working on an incremental refactoring of the configuration of this addon. It makes sense to ship all changes in one release to avoid unnecessary noice.

I would suggest that next release includes these features:

* [x] Move configuration to `ember-cli-content-security-policy` namespace, introduce an explicit `enabled` option and reword other options. (#92, #94)
* [x] Move configuration to `config/content-security-policy.js`. (#30, #97, #104)
* [x] Policy object should be replaced by application configuration not merged. (#99, #101)

There are some pending feature requests that could be implemented having sensitive defaults without introducing a breaking change if shiped together with configuration changes:

* [x] Add an assertion to fail tests if CSP is violated. (#91)
* [x] Add CSP header in FastBoot if addon is enabled and `delivery` includes `header`. (#113)

I suggest shipping all of them with the next release.

Also there are some bugs/regressions introduced by refactoring that must be fixed before release:

* [x] CLI command is not using refactored configuration. (#98, #103)
* [x] Runtime configuration should not be injected if not needed (#116, #121)
* [x] Throws in FastBoot if disabled or delivery does not include header (#117, #118)
* [x] Ensure consistent test results by always using test environment for CSP affecting tests (#119, #122)

Steps to be done after release:

* [ ] Update Ember CLI docs for [Content Security Policy](https://cli.emberjs.com/release/basic-use/deploying/#contentsecuritypolicy) (https://github.com/ember-learn/cli-guides/pull/146)

This would be a great foundation for https://github.com/jelhan/ember-style-modifier/issues/11.

sandstrom:
Sounds good! Ping me or rwjblue if you want us to review or merge something. 

You can also ask rwjblue for repo access, perhaps he can give you. However, even if you'd have repo access I still think it would be vice to have another pair of eyes look through code before it's merged.

rwjblue:
OK, just merged #101 which I think ticks a couple of the boxes above.

jelhan:
[...]

PR #101, opened 2019-07-03
Avoid merging policies in build time configuration

### LINKED ISSUE #99: Policy object should not be merged but replaced by application configuration
Currently the default policy object is merged with application configuration. This is confusing and is likely causing misconfiguation.

Lets assume someone tries to forbid any external assets. So he aims for a CSP Header as following:

```http
Content-Security-Policy: default-src 'none';
```

He would most likely try a configuration as following:

```js
let app = new EmberApp(defaults, {
  'ember-cli-content-security-policy': {
     policy: {
       'default-src': ["'none'"]
     }
  }
});
```

But that would result in the following CSP Header as it's merged with default policy object:

```
Content-Security-Policy: default-src 'none'; script-src 'self'; font-src 'self'; connect-src 'self'; img-src 'self' data:; style-src 'self'; media-src 'none';
```

In order to achieve the goal this configuration is needed:

```js
let app = new EmberApp(defaults, {
  'ember-cli-content-security-policy': {
     policy: {
       'default-src': ["'none'"],
       'script-src':  null,
       'font-src':    null,
       'connect-src': null,
       'img-src':     null,
       'style-src':   null,
       'media-src':   null,
     }
  }
});
```

This is mentioned in README but easy to be missed:

> To clear a directive from the default policy, set it to `null`.

The general refactoring of configuration done in #94 and #97 is a perfect time to change this in my opinion.

Having the policy object been merged might not have been a clear decision but might likely be caused by implementation details of Ember CLI's `config` hook.

Please note that the merging described here is currently broken in master due to #94. It will be fixed with #97, which also adds tests for this one. We would still need to support it for legacy run time configuration option `contentSecurityPolicy` to not add a breaking change.

jelhan:
Should be backward-compatible as it still merges the policy object set by `contentSecurityPolicy` runtime config.

Closes #99

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,16 +271,7 @@ function calculateConfig(environment, buildConfig, runConfig, ui) {
   }
 
   // apply configuration
-  // policy object is merged not replaced
-  if (buildConfig[CONFIG_KEY]) {
-    Object.keys(buildConfig[CONFIG_KEY]).forEach((key) => {
-      if (key === 'policy') {
-        Object.assign(config.policy, buildConfig[CONFIG_KEY].policy);
-      } else {
-        config[key] = buildConfig[CONFIG_KEY][key];
-      }
-    });
-  }
+  Object.assign(config, buildConfig[CONFIG_KEY]);
 
   return config;
 }
--- a/node-tests/unit/config-test.js
+++ b/node-tests/unit/config-test.js
@@ -25,12 +25,13 @@ describe('unit: configuration', function() {
     expect(config.reportOnly).to.be.true;
   });
 
-  it('merges policy object with default one', function() {
+  it('replaces default policy object with application config', function() {
     let config = calculateConfig(
       'development',
       {
         'ember-cli-content-security-policy': {
           policy: {
+            'default-src': ["'self'"],
             'font-src': ['examples.com']
           }
         }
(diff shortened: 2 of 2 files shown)


jelhan:
@sandstrom This is ready to be reviewed and merged from my side.

jelhan:
@rwjblue Seems like sandstorm is busy. Would be great if you have some time for a review.

jelhan:
> LGTM, this will be a major bump, right (since it changes merging behaviors)?

It only changes merging behavior for policies that are defined using new configuration but not for legacy configuration. There wasn't any release yet that includes new configuration. So it should be backward compatible. There is also a test that legacy configuration is merged, which isn't touched by this PR: https://github.com/rwjblue/ember-cli-content-security-policy/blob/2266d1a7bba293af4330459b0ffb9c94a3042fee/node-tests/unit/config-test.js#L53-L74

jelhan:
Please have a look at #100 for a proposal for next release. The main idea is to use the altered configuration interface as a backward-compatible toggle for new defaults. It lists the changes that should be bundled together to prevent the need of a breaking change.

rwjblue:
> It only changes merging behavior for policies that are defined using new configuration but not for legacy configuration. There wasn't any release yet that includes new configuration. So it should be backward compatible.

Awesome, thank you. I'm sorry I missed that.

PR #103, opened 2019-07-25
CLI command should respect new configuration

### LINKED ISSUE #98: CLI command is not using refactored configuration
`ember csp-headers` is not using refactored configuration. Needs to be updated accordingly to #94 and #97. Should reuse `calculateConfig` function introduced in #97 to avoid similar issues.

This bug only affects current `master` as there wasn't a release since #94.

jelhan:
Closes #98

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -2,8 +2,15 @@
 
 'use strict';
 
-var buildPolicyString = require('./utils')['buildPolicyString'];
-var chalk = require('chalk');
+const chalk = require('chalk');
+const {
+  buildPolicyString,
+  calculateConfig,
+  readConfig
+} = require('./utils');
+
+const CSP_HEADER              = 'Content-Security-Policy';
+const CSP_HEADER_REPORT_ONLY  = 'Content-Security-Policy-Report-Only';
 
 module.exports = {
   'csp-headers': {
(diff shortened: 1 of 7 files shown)


lib/commands.js
```diff
@@ -31,22 +39,28 @@ module.exports = {
     ],
 
     run: function(options) {
-      var config = this.project.config(options.environment);
-      var reportUri = options.reportUri;
+      let { environment, reportUri, silent } = options;
+      // TODO: find a way to get the build time configuration in Ember CLI command
+      let buildConfig = {};
```
jelhan: @rwjblue I would be glad for any hint how to get build time configuration in an Ember CLI command. I didn't found any way. :cry:

jelhan: What I've tried so far:

- Read the documentation of Ember CLI command and debugged `this`. I wasn't able to find any reference to app's build time configuration.
- Done a code search on EmberObserver for `includedCommands` and digged through the results. Wasn't able to find any addon that consumes app's build time configuration in an Ember CLI command.
- Construct path to app's `ember-cli-build.js` using `this.project.root` and pass it to [`ember-cli-build-config-editor`](https://github.com/srvance/ember-cli-build-config-editor). This one parses the file (AST) and provides a method to retrieve the configuration. But it does not support complex data structures like objects or arrays as config values. Therefore we can't use it as we have a very complex data structure for `policy` config option.
- Require app's `ember-cli-build.js` and initialize an instance of the app. This app should have the build time configuration as `options` property but it's always empty for some reason. This seems to depend on the `defaults` argument passed to the constructor - and I don't know how to mock that one.

I've no idea what options are left. Seems like a show stopper to me.

rwjblue: > Require app's ember-cli-build.js and initialize an instance of the app. This app should have the build time configuration as options property but it's always empty for some reason. This seems to depend on the defaults argument passed to the constructor - and I don't know how to mock that one.

I think this is the right path for us at the moment. The argument passed and referred to as `defaults` is _essentially_ `{ project }` ([see here](https://github.com/ember-cli/ember-cli/blob/3575b9b8a96a392a20ae138290dd085667b579a7/lib/models/builder.js#L52)). 

Alternatively, we could move our config into a specific path that we can require (e.g. `config/content-security-policy.js`) instead of using `ember-cli-build.js` directly.

jelhan: I don't think initializing an instance of the app is an option here. I've investigated that one further and noticed that `ember-cli-build.js` returns a broccoli tree. It's `options` property is not populated with app's build options but with options passed to `broccoli-merge-trees`. Therefore it's an empty hash.

> Alternatively, we could move our config into a specific path that we can require (e.g. `config/content-security-policy.js`) instead of using `ember-cli-build.js` directly.

 This seems to be the path forward to me. Will implement it in another PR.

rwjblue: Sounds good. I agree this seems like the most sensible option...

jelhan: @rwjblue Please review #104 for implementation.

jelhan:
Rebased on top of #104. Tests should pass now. Still a WIP as #104 needs to be merged before. Will introduce merge conflicts with #105. I don't care which one is merged before.

jelhan:
Rebased on top of #105. Should be ready as soon as that one is merged.

jelhan:
Rebased on top of master. @rwjblue ready to be merged from my side.

jelhan:
Rewritten commit history so that #91 is not blocked anymore by this one but shares a commit with it.

rwjblue:
Sorry @jelhan, I think this will need a rebase (on top of #107). Other than that, this looks good to me...

jelhan:
@rwjblue Rebased. Not sure what happened with Travis CI. Seems like it created five builds. Four seem to be stale. But should be safe to merge cause one passed.

lib/commands.js
```diff
@@ -2,8 +2,15 @@
...
-var buildPolicyString = require('./utils')['buildPolicyString'];
-var chalk = require('chalk');
+const chalk = require('chalk');
+const {
+  buildPolicyString,
+  calculateConfig,
+  readConfig
+} = require('./utils');
+
+const CSP_HEADER              = 'Content-Security-Policy';
```
rwjblue: Definitely non-blocking, and probably just a bit too nitpicky but _in general_ I prefer to avoid the equals alignment stuff. It makes future changes change lines that are unrelated (e.g. when adding new variables with a slightly longer name), and then makes it just a tiny bit harder to follow whats going on through the git history.

jelhan: :100: I was just adopting the existing coding style.

PR #104, opened 2019-08-04
move config to config/content-security-policy.js

jelhan:
Moves configuration from `ember-cli-build.js` to `config/content-security-policy.js`.

This is mainly motivated by the complexity required to get build-time configuration in different stages of an addon. Build-time config is not available in all hooks and does not seem to be available at all for commands.

This should unblock #103.


jelhan:
@sandstrom This was discussed with rwjblue in https://github.com/rwjblue/ember-cli-content-security-policy/pull/103#discussion_r307498314 cause otherwise the configuration is not accessible from Ember CLI commands. Please raise any concerns as this changes what we had discussed in #92.

lib/utils.js
```diff
@@ -22,4 +26,20 @@ var buildPolicyString = function(policyObject) {
...
-module.exports = { buildPolicyString: buildPolicyString };
+/**
+ * Returns the configuration stored in `config/content-security-policy.js`.
+ * Returns an empty object if that file does not exist.
+ *
+ * @param {string} projectRoot
+ * @return {object}
+ */
+const readConfig = function(projectRoot, environment) {
+  let configPath = path.join(projectRoot, 'config/content-security-policy.js');
```
rwjblue: This logic needs to be tweaked a bit, the `config` folder isn't always in a static location.

[Here is a similar utility](https://github.com/emberjs/ember-optional-features/blob/410c8dc0e2273e4d0f538c0e233a9ec659c5e291/utils.js#L3-L11) I built a little while back, but the tldr is _basically_:

```js
function getConfigPath(projectPkg, projectRoot) {
  let configDir = 'config';

  if (projectPkg['ember-addon'] && projectPkg['ember-addon']['configPath']) {
    configDir = projectPkg['ember-addon']['configPath'];
  }

  return path.join(projectRoot, configDir, 'content-security-policy.js');
}
```

For addon's that depend on ember-cli-content-security-policy the config path would then be `tests/dummy/config/content-security-policy.js` (which allows them to use this addon for their deployed demo app for example), other projects need to use the `config` folder for "something else" (don't ask :weary:) and therefore customize `ember-addon.configPath` to be `ember-cli-config`.

jelhan:
@rwjblue Thanks for merging that quickly. Actually I had planed to update the README as suggested but hasn't finished yet. You could find that one in #106.

rwjblue:
Whoops, sorry about that! Seems good to be separate too though.

PR #105, opened 2019-08-05
fix livereload

jelhan:
Allowing connections and script source required by livereload feature was broken by configuration refactoring. This fixes it and adds tests.

This PR is based on #104. It should be merged afterwards.

### DIFF
--- a/index.js
+++ b/index.js
@@ -71,8 +71,8 @@ let appendSourceList = function(policyObject, name, sourceList) {
 };
 
 // appends directives needed for Ember CLI live reload feature to policy object
-let allowLiveReload = function(policyObject) {
-  let { hostname, port, ssl } = this._config.liveReload;
+let allowLiveReload = function(policyObject, liveReloadConfig) {
+  let { hostname, port, ssl } = liveReloadConfig;
 
   ['localhost', '0.0.0.0', hostname].filter(Boolean).forEach(function(hostname) {
     let protocol = ssl ? 'wss://' : 'ws://';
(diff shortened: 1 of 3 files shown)


rwjblue:
Ready for a rebase now that #104 has landed

jelhan:
Rebased and fixed the test that were failing due to leaking state.

jelhan:
This is not ready yet. This had moved the calculation of config to `serverMiddleware` hook but that's only executed on `ember serve` but not on `ember test` and `ember build`. Will need to refactor to calculate the config in a more common hook (e.g. `included` as before) and handle the livereload configuration separately.

jelhan:
Pushed a new implementation. It's a total rewrite of this PR. I guess a new review + approval is required.

It does not treat live reload options as part of config anymore. They are calculated independently from other config in `serverMiddleware` hook and cached for `contenFor` hook. Live reload should be disabled if `serverMiddleware` hook is not executed cause live reload depends on express server. Luckily `contentFor` hook is executed after `serverMiddleware` hook.

Before my rewrites of all that stuff live reload was calculated in `contentFor` and `serverMiddleware` hooks independently:
https://github.com/rwjblue/ember-cli-content-security-policy/blob/4aabfa50496c720a76fa1e689ea43c04a40be875/index.js#L122-L129
https://github.com/rwjblue/ember-cli-content-security-policy/blob/4aabfa50496c720a76fa1e689ea43c04a40be875/index.js#L173-L184

This introduces inconsistency between both which may result in broken live reload support if CSP is put in meta tag. Especially the `contentFor` hook ignored the live reload host option cause that value wasn't available trough environment variables.

jelhan:
Squashed commits.

PR #106, opened 2019-08-06
upgrade readme

jelhan:
As suggested in #104.

Please have a close look at TypeScript definition as I'm not that used to write TypeScript yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,36 +23,61 @@ Installation
 ember install ember-cli-content-security-policy
 ```
 
-## Configuration
+Configuration
+------------------------------------------------------------------------------
 
 This addon is configured via `config/content-security-policy.js` file.
 
-- `delivery: string[]`
-  CSP is delivered via HTTP Header if delivery includes `"header"` and via meta element if it includes `"meta"`.
-  Defaults to `["header"]`.
-- `enabled: boolean`
-  Controls if addon is enabled at all.
-  Defaults to `true`.
-- `policy: object`
-  A hash of options representing a Content Security Policy.
-  Defaults to:
-  ```js
-  {
-    'default-src':  ["'none'"],
-    'script-src':   ["'self'"],
-    'font-src':     ["'self'"],
-    'connect-src':  ["'self'"],
-    'img-src':      ["'self'"],
-    'style-src':    ["'self'"],
-    'media-src':    ["'self'"],
-  }
-  ```
-  To clear a directive from the default policy, set it to `null`.
-  The browser will fallback to the `default-src` if a directive does not exist.
-- `reportOnly: boolean`
-  Controls if CSP is used in report only mode. For delivery mode `"header"` this causes `Content-Security-Policy-Report-Only` HTTP header to be used.
-  Can not be used together with delivery mode `"meta"` as this is not supported by CSP spec.
-  Defaults to `true`.
+```ts
+interface EmberCLIContentSecurityPolicyConfig {
+  // CSP is delivered via HTTP Header if delivery includes `"header"` and via
+  // meta element if it includes `"meta"`.
+  delivery?: string,
+
+  // Controls if addon is enabled at all.
+  enabled?: boolean,
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,36 +23,61 @@ Installation
...
-    'default-src':  ["'none'"],
-    'script-src':   ["'self'"],
-    'font-src':     ["'self'"],
-    'connect-src':  ["'self'"],
-    'img-src':      ["'self'"],
-    'style-src':    ["'self'"],
-    'media-src':    ["'self'"],
-  }
-  ```
-  To clear a directive from the default policy, set it to `null`.
```
jelhan: Since the configuration is not merged anymore, there isn't any need to set a directive to `null`. Missed to delete that sentence in #101.

README.md
```diff
@@ -23,36 +23,61 @@ Installation
...
+  // A hash of options representing a Content Security Policy. The key must be
+  // a CSP directive name as defined by spec. The value must be an array of
+  // strings that form a CSP directive value, most likely a source list, e.g.
+  // {
+  //   'default-src': ["'none'"],
+  //   'style-src': ["'self'", 'examples.com']
+  // }
+  // Please refer to CSP specification for details on valid CSP directives:
+  // https://w3c.github.io/webappsec-csp/#framework-directives
+  policy?: { [key: string]: string[]; },
```
jelhan: May provide a list of allowed keys but that would quickly get outdated as new keys might be added by CSP3, which is still an editor's draft or even by other specs ([§6.5 Directives Defined in Other Documents](https://w3c.github.io/webappsec-csp/#directives-elsewhere)).

README.md
```diff
@@ -23,36 +23,61 @@ Installation
...
-  The browser will fallback to the `default-src` if a directive does not exist.
-- `reportOnly: boolean`
-  Controls if CSP is used in report only mode. For delivery mode `"header"` this causes `Content-Security-Policy-Report-Only` HTTP header to be used.
-  Can not be used together with delivery mode `"meta"` as this is not supported by CSP spec.
-  Defaults to `true`.
+```ts
+interface EmberCLIContentSecurityPolicyConfig {
+  // CSP is delivered via HTTP Header if delivery includes `"header"` and via
+  // meta element if it includes `"meta"`.
+  delivery?: string,
```
rwjblue:
```suggestion
  delivery?: 'meta' | 'header',
```

README.md
rwjblue: TS interfaces allow additional keys, so I think we can list the ones we expect at the moment (AFAIK new keys are still _somewhat_ uncommon)

jelhan: Added the directive names as a type referenced as key. I hope that was the correct way. It's a very long list. But it's maybe helpful to explicitly state that all directives are supported and not only that one used in default config.

jelhan:
@rwjblue Thanks for quick review! Addressed the feedback.

jelhan:
Is there any way we could expose that type information to be used by consumers? Would provide helpful auto-completion and prevent typos in directive names.

rwjblue:
> Is there any way we could expose that type information to be used by consumers? Would provide helpful auto-completion and prevent typos in directive names.

Honestly, I'm not 100% sure but I think we could publish a .d.ts file so folks that wanted to use .ts could type their `config/content-security-policy.js`:

```ts
module.exports = function(environment: string): import('ember-cli-content-security-policy').PolicyConfiguration {
  // return stuff....
}
```

jelhan:
> Honestly, I'm not 100% sure but I think we could publish a .d.ts file so folks that wanted to use .ts could type their `config/content-security-policy.ts`:
> 
> ```ts
> module.exports = function(environment: string): import('ember-cli-content-security-policy').PolicyConfiguration {
>   // return stuff....
> }
> ```

I will have a look into that one. Maybe some other addon is already doing it. Will have a look through possible candidates.

This PR is ready to be merged from my side.

PR #107, opened 2019-08-08
Ensure `csp-headers` command emits to standard out (to allow for piping into other programs)

Exelord:
1. `this.ui.writeLine` is automatically covering the `--silent` option. There is no need to wrap it up with the if.

2. The result should be always logged even in silent mode. There is no point to run this command without returning a header payload. That's why I used `console.log` as it logs always. Also, this will allow for saving the result to file eg: `ember csp-headers --silent > csp.txt` which can be later used during the deploy etc...

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -38,15 +38,14 @@ module.exports = {
         config.contentSecurityPolicy['report-uri'] = reportUri;
       }
 
-      if (!options.silent) {
-        this.ui.writeLine(chalk.dim.cyan('# Content Security Policy Header Configuration'));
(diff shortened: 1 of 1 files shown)


jelhan:
This moves the output from `stderr` to `stdout`. I'm not sure if this should be considered a breaking change or a bug fix. It will likely break existing scripts that uses the output.

rwjblue:
> It will likely break existing scripts that uses the output.

Ya, agree. Though I _think_ we are already queuing master up to be a breaking change release (due to things like Node version support changes), right?

Issue #67, opened 2016-07-08
hashed unsafe-inline support?


jamesarosen:
`unsafe-inline` isn't sufficient to allow `<script>...</script>` on the page. CSP 2 requires those tags to have a hash of their contents. See https://www.w3.org/TR/2015/CR-CSP2-20150721/#script-src-hash-usage

Is there a way for this library to automatically calculate the hashes? Or should addons that use `contentFor` do the hashing and add the results to `config.contentSecurityPolicy['script-src']`? If the latter, could this addon expose an API to make that easier?

sandstrom:
Thanks for bringing this up James!

At the moment no, but I agree that it would be a good addition.

For hashes we'd need to have this addon run after minification. Do you know if ember-cli has support for specifying the order of addons?

I don't know much about addon <-> addon communication, but unless there is a way of specifying the order it would be better if this addon had an API for setting the hash.

***

Nonces (https://www.w3.org/TR/2015/CR-CSP2-20150721/#script-src-nonce-usage) would be easier to implement, but won't work since they must be randomly generated for each request (ember is often hosted statically).

jelhan:
This is also a requirement for `ember-style-modifier` to support SSR/Fastboot: https://github.com/jelhan/ember-style-modifier/issues/11 Please especially note the design outlined in https://github.com/jelhan/ember-style-modifier/issues/11#issuecomment-499795602. I'm planing to implement that feature soon.

GCheung55:
Any update on this issue? I would be really helpful to have a way to add hashes to the configuration at build time.

jelhan:
It's still a work in progress. But slow progress to be honest. v2 should lay the foundation for this feature. I hope that implementation is straight forward afterwards. But I don't have much time to work on this addon currently. So it will take some time.

I would appreciate any help. If you have some time to help implementing this feature or the missing pieces to get v2 stable, please reach out to me on Discord to coordinate.

PR #113, opened 2019-09-16
set CSP header in FastBoot

jelhan:
Let me give a quick overview of the architecture. Especially cause I'm not that used to FastBoot development and there might be problems that I have missed.

- Addon provides part of it's build-time configuration (`reportOnly` and the calculated policy string) as run-time configuration by returning it from `config` hook. Therefore the calculation was moved from `included` hook to `config` hook.
- It provides an instance-initializer, which runs only in FastBoot. That one reads the run-time configuration and sets the CSP header using the `fastboot` service.
- If running `ember serve` not only FastBoot but also the registered express middleware in `serverMiddleware` hook sets CSP header. Test is using `fastboot-app-server` to prevent false-positives.

There are some potential improvements, which would reduce bundle size:

- Run-time configuration should only be provided if addon has `ember-cli-fastboot` as a dependency.
- There is no need to provide the policy string through run-time configuration if it's already available as CSP meta tag.

Documented that ones as todos in code. To be honest I don't think that few bytes would make a real impact on performance. :innocent: 

This lays the foundation for manipulation of the policy per request as required by #67 and https://github.com/jelhan/ember-style-modifier/issues/11.


jelhan:
Need to check what is going on with the tests. They seem to be flickering. Added _WIP_. Sorry for the noise.

addon/instance-initializers/content-security-policy.js
```diff
@@ -0,0 +1,34 @@
+import { assert } from '@ember/debug';
+
+// reads addon config stored in meta element
+function readAddonConfig(appInstance) {
+  let config = appInstance.resolveRegistration('config:environment');
```
rwjblue: Doesn't this read the config from the "legacy" location (in `config/environment.js`)?

addon/instance-initializers/content-security-policy.js
```diff
@@ -0,0 +1,34 @@
...
+  assert(
+    'Required configuration is available at run-time',
+    addonConfig.hasOwnProperty('reportOnly') && addonConfig.hasOwnProperty('policy')
+  );
+
+  return config['ember-cli-content-security-policy'];
+}
+
+export function initialize(appInstance) {
+  let fastboot = appInstance.lookup('service:fastboot');
```
rwjblue: Can you wrap the guts of this function in:

```js
if (typeof FastBoot !== 'undefined') {

}
```

That is probably the fastest way to bail out in that circumstance...

addon/instance-initializers/content-security-policy.js
jelhan: You are right that it reads the run-time configuration, which is normally provided by consumer via `config/environment.js`. But in this case the configuration is meant to be provided by addon's `config` hook only. It's a subset of addons configuration. It only includes `reportOnly` option and the build policy string. Both are required at run-time for FastBoot support.

package.json
```diff
@@ -19,11 +19,12 @@
     "start": "ember serve",
     "test": "ember test",
     "test:all": "ember try:each",
-    "test:node": "mocha node-tests/**"
+    "test:node": "for i in node-tests/*/*; do mocha $i; done"
```
jelhan: There seems to be strange issues with `ember-cli-addon-tests` triggered only if multiple tests are executed at the same time and also depending on the order in which the tests are executed. While this is definitely not a good solution it gets us unblocked here.

addon/instance-initializers/content-security-policy.js
jelhan: Are you sure this is needed? I thought putting the instance initializer in `fastboot/` directory ensures that will only be included if app uses FastBoot.

jelhan:
This is ready to be merged if we accept the ugly hack for CI. There is one question still open, which I would love to get some feedback.

>  Should there be an option to disable setting CSP headers in FastBoot?

I'm tending towards no, cause we have the `enabled` and `delievery` options. Both could be used to not set headers in FastBoot. I'm not sure if I miss a scenario in which development server should set CSP headers but FastBoot should not.

@rwjblue @sandstrom Do you have any input on that one?

sandstrom:
@jelhan I don't think it makes sense with dedicated disabling of CSP for Fastboot. If that need arises, one could simply use the existing `enabled` option to disable.

package.json
jelhan: Reported this upstream: https://github.com/tomdale/ember-cli-addon-tests/issues/215

jelhan:
@rwjblue Any objections merging this one?

fastboot/instance-initializers/content-security-policy.js
```diff
@@ -0,0 +1 @@
+export { default, initialize } from 'ember-cli-content-security-policy/instance-initializers/content-security-policy';
```
rwjblue: Can we move the actual implementation of the initializer here instead? Specifically, I'd like to avoid adding the initializer to the build output for users that _aren't_ using fastboot.

In general, I definitely prefer what you've done here (having the actual implementation inside the `addon/` folder), but if we don't move that file here we will force _every_ user to have that extra module...

package.json
rwjblue: Eeck, seems fine but that sounds like it was pretty gnarly to track down.

fastboot/instance-initializers/content-security-policy.js
jelhan: Done in 1fe117fb4adce636c4394c433698eaed8ef3044b.

jelhan:
Added some documentation in 42b1e83321e382be3e34b27751d3fb622dfbfa1d. There isn't much to say about the default usage as this should just work.

index.js
```diff
@@ -49,6 +49,55 @@ let allowLiveReload = function(policyObject, liveReloadConfig) {
...
+        appendSourceList(config.policy, 'script-src', `'nonce-${STATIC_TEST_NONCE}'`);
+      }
+
+      this._config = config;
+      this._policyString = buildPolicyString(config.policy);
+    }
+
+    // provide configuration needed at run-time for FastBoot support (if needed)
+    // TODO: only inject if application uses FastBoot
+    if (!this._config.enabled || !this._config.delivery.includes('header')) {
```
rwjblue: How does the delivery mode get set to header? Should have a specific opt-in for the FastBoot behavior (via either a special delivery mode, or something)?

index.js
```diff
@@ -49,6 +49,55 @@ let allowLiveReload = function(policyObject, liveReloadConfig) {
...
+      if (app && app.tests) {
+        appendSourceList(config.policy, 'script-src', `'nonce-${STATIC_TEST_NONCE}'`);
+      }
+
+      this._config = config;
+      this._policyString = buildPolicyString(config.policy);
+    }
+
+    // provide configuration needed at run-time for FastBoot support (if needed)
+    // TODO: only inject if application uses FastBoot
```
rwjblue: Would you mind making an issue for this, and linking to the issue from the TODO comment? I'm mostly just trying to make sure we don't loose track of this and that we know when the TODO can be removed...

index.js
rwjblue: Also, should we make a constant for `DELIVERY_HEADER` (or change the usage of `DELIVERY_META` to just be `'meta'`, which seems better to me personally)? It's a tad odd that we treat them differently in different places...

jelhan: > How does the delivery mode get set to header?

This is the `delivery` configuration option. It's set via `config/content-security-policy.js`. It defaults to `["header"]`.

> Should have a specific opt-in for the FastBoot behavior (via either a special delivery mode, or something)?

I'm not sure. That's the open question in PR text. I tend to believe that `enabled` and `delivery` method gives enough control to disable setting for FastBoot. One might either remove `"header"` from `delivery` or disable the addon at all in production environment. I can't image any use case in which you need `delivery` to include `"header"` but don't want the header to be set in FastBoot. Beside FastBoot support the delivery method `"header"` only controls if CSP headers are set by Ember CLI development server. If I got @sandstrom right, he has the same feeling (https://github.com/rwjblue/ember-cli-content-security-policy/pull/113#issuecomment-536521562).

jelhan: > Also, should we make a constant for `DELIVERY_HEADER` (or change the usage of `DELIVERY_META` to just be `'meta'`, which seems better to me personally)? It's a tad odd that we treat them differently in different places...

Removed `DELIVERY_META` and `DELIVERY_HEADER` consts in b118a5dfc19b2baea194f8c7df389b192ac5b5a5.

index.js
jelhan: Created #116 and named that one in source code.

rwjblue: Thank you!

index.js
jelhan: I would not mix that both concerns. There is also a delivery option `"meta"`, which causes the policy to be injected as a meta tag. This is totally valid concern regardless of FastBoot.

If the addon is enabled in production, consuming app does not use FastBoot and delivery does *not* include `"meta"`, the addon should not have any effect. If you are concerned about unnecessary run-time config, I would rather fix #116 before releasing a new version than having complex default due to it. Fixing #116 should not be more difficult than having defaults depending on the presence of ember-cli-fastboot.

rwjblue: > If you are concerned about unnecessary run-time config

Yes, that is exactly my concern. My comment above should likely have _gone on #116_. 😸

PR #115, opened 2019-10-06
WIP: run tests against different Ember CLI versions in CI

jelhan:


rwjblue:
> If it's okay for you, I would reduce support matrix to v3.4 or above. This would cover three LTS versions.

Seems good to me. 3.4, 3.8, and 3.12 sound perfect to me...

PR #118, opened 2019-10-10
inject instance initializer only if required

### LINKED ISSUE #117: Throws in FastBoot if disabled or delivery does not include header
The run-time configuration is only present if addon is enabled and delivery contains `"header"`: https://github.com/rwjblue/ember-cli-content-security-policy/blob/7c7031fc2984519f567a3fb6643ceb63bab05148/index.js#L85-L90

This is by design to prevent unnecessary bloat.

This run-time configuration is consumed by an instance initializer which is only present if the application depends on `ember-cli-fastboot`. It assert that the run-time configuration is present: https://github.com/rwjblue/ember-cli-content-security-policy/blob/7c7031fc2984519f567a3fb6643ceb63bab05148/fastboot/instance-initializers/content-security-policy.js#L10-L13 

If that assertion is stripped in a production build, the next line will throw with a `TypeError` cause `config` is undefined: https://github.com/rwjblue/ember-cli-content-security-policy/blob/7c7031fc2984519f567a3fb6643ceb63bab05148/fastboot/instance-initializers/content-security-policy.js#L15

The instance initializer is present even if addon is disabled or delivery does not include `"header"`. Since the run-time configuration is not available, it will throw.

This should be fixed by only including the instance initalizier in the build if required. This would also prevent us from pushing code to the consuming application that is not needed.

jelhan:
Removes `fastboot/instance-initializers/content-security-policy.js` from tree if CSP headers should not be set by FastBoot. This prevents unnecessary code in build output. As the same scenarios are used as for not injecting the run-time configuration it fixes #117 also.

It also increases the test coverage. Especially it ensures that there wasn't an error thrown at FastBoot render and verifies that run-time configuration is not part of the build.

### DIFF
--- a/index.js
+++ b/index.js
@@ -82,13 +82,24 @@ module.exports = {
       this._policyString = buildPolicyString(config.policy);
     }
 
-    // provide configuration needed at run-time for FastBoot support (if needed)
-    // TODO: only inject if application uses FastBoot
-    //       https://github.com/rwjblue/ember-cli-content-security-policy/issues/116
-    if (!this._config.enabled || !this._config.delivery.includes('header')) {
+    // CSP header should only be set in FastBoot if
+    // - addon is enabled and
+    // - configured to deliver CSP via header and
+    // - application has ember-cli-fastboot dependency.
+    this._needsFastBootSupport = this._config.enabled &&
+      this._config.delivery.includes('header') &&
+      // TODO: check if application has ember-cli-fastboot-dependency
+      // https://github.com/rwjblue/ember-cli-content-security-policy/issues/116
+      true;
+
+    // Run-time configuration is only needed for FastBoot support.
+    if (!this._needsFastBootSupport) {
       return {};
     }
 
+    // In order to set the correct CSP headers in FastBoot only a limited part of
+    // configuration is required: The policy string, which is used as header value,
+    // and the report only flag, which is determines the header name.
     return {
       'ember-cli-content-security-policy': {
         policy: this._policyString,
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -82,13 +83,24 @@ module.exports = {
...
-    if (!this._config.enabled || !this._config.delivery.includes('header')) {
+    // CSP header should only be set in FastBoot if
+    // - addon is enabled and
+    // - configured to deliver CSP via header and
+    // - application has ember-cli-fastboot dependency.
+    this._needsFastBootSupport = this._config.enabled &&
+      this._config.delivery.includes('header') &&
+      // TODO: check if application has ember-cli-fastboot-dependency
+      // https://github.com/rwjblue/ember-cli-content-security-policy/issues/116
+      true;
```
jelhan: Using `ember-cli-version-checker` to check if `ember-cli-fastboot` exists should be straight forward here. Actually I was working on that one before facing #117. But I like to do it in another PR to not overcomplex this one. Also I fear that I will face testing issues caused by `ember-cli-version-checker` and `ember-cli-addon-tests` not playing well together...

index.js
```diff
@@ -221,9 +233,25 @@ module.exports = {
     return require('./lib/commands');
   },
 
+  treeForFastBoot: function(tree) {
+    // Instance initializer should only be included in build if required.
+    // It's only required for FastBoot support.
+    if (!this._needsFastBootSupport) {
+      tree = funnel(tree, {
```
rwjblue: The tree argument passed in here is our own `./fastboot` folder, which should only contain the excluded file. I think we should change this to `return null` instead of excluding the specific file...

jelhan: Done so in 32e5d207490a6b5259b0454fdbf4b2790905b356.

jelhan:
Addressed review comments and squashed the commits afterwards to prevent an ugly history.

PR #121, opened 2019-12-15
inject runtime config only if FastBoot dependency exists

### LINKED ISSUE #116: Runtime configuration should not be injected if application does not have FastBoot dependency
#113 added some run-time configuration that is required for FastBoot support only. This configuration should not be injected in app's runtime configuration if the app does not have a dependency on FastBoot: https://github.com/rwjblue/ember-cli-content-security-policy/blob/42b1e83321e382be3e34b27751d3fb622dfbfa1d/index.js#L87-L91

jelhan:
Ensures that runtime configuration is not injected if it's not needed cause consuming app does not have a dependency on FastBoot.

Have verified manually that it's working by using this feature branch in a new app. `ember-cli-version-checker` seems to report wrong information if used with `yarn link` and `ember-cli-content-security-policy`, which makes testing this one difficult and may cause confusion. Symlinking the addon may be causing the issue in both cases. Need to investigate a little bit more and report their.

It also sounds similar to https://github.com/ember-cli/ember-cli-version-checker/issues/93. There the issue was traced down that some dependencies was causing an old version of ember-cli-version-checker to be used. Need to investigate if that one cause the issues here as well. But will do separately. Please feel free to merge this one with the skipped test.

Closes #116

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,9 +88,7 @@ module.exports = {
     // - application has ember-cli-fastboot dependency.
     this._needsFastBootSupport = this._config.enabled &&
       this._config.delivery.includes('header') &&
-      // TODO: check if application has ember-cli-fastboot-dependency
-      // https://github.com/rwjblue/ember-cli-content-security-policy/issues/116
-      true;
+      new VersionChecker(this).for('ember-cli-fastboot').exists();
 
     // Run-time configuration is only needed for FastBoot support.
     if (!this._needsFastBootSupport) {
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@ module.exports = {
     // - application has ember-cli-fastboot dependency.
     this._needsFastBootSupport = this._config.enabled &&
       this._config.delivery.includes('header') &&
-      new VersionChecker(this).for('ember-cli-fastboot').exists();
+      new VersionChecker(this.project).for('ember-cli-fastboot').exists();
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -88,9 +88,7 @@ module.exports = {
     // - application has ember-cli-fastboot dependency.
     this._needsFastBootSupport = this._config.enabled &&
       this._config.delivery.includes('header') &&
-      // TODO: check if application has ember-cli-fastboot-dependency
-      // https://github.com/rwjblue/ember-cli-content-security-policy/issues/116
-      true;
+      new VersionChecker(this).for('ember-cli-fastboot').exists();
```
rwjblue: Changing this should (at least in theory) help identify if the _project_ has `ember-cli-fastboot` installed, and may help the linking scenario to allow us to unskip it.

```suggestion
      new VersionChecker(this.project).for('ember-cli-fastboot').exists();
```

jelhan: Thanks a lot for pushing me in the right direction! It's working like a charm. Will fix this also for the QUnit warning in a follow up PR: https://github.com/rwjblue/ember-cli-content-security-policy/blob/f985629287f3466b02fcd94edda84f0b493595cc/index.js#L236 https://github.com/rwjblue/ember-cli-content-security-policy/blob/f985629287f3466b02fcd94edda84f0b493595cc/node-tests/e2e/test-support-test.js#L195

PR #122, opened 2019-12-20
consistent test results regardless of environment

### LINKED ISSUE #119: Ensure consistent test results by always using test environment for CSP affecting tests
Tests are affected by CSP. Code may be blocked by a policy or the test may even fail due to test support integration coming with this addon. In the current implementation different policies may be used for tests depending on the way the tests are started:

1. If run with `ember test` or `ember test --server` the CSP for `test` environment is used. A delivery through meta tag is enforced to ensure that CSP is applied. This is necessary cause `ember test` does not use built-in webserver.
2. If run with `ember server` and opening `http://locahost:4200/tests` in a browser `development` environment is used. If `--environment` option is passed to `ember serve` any environment may be used.
3. If build is not served using Ember CLI and tests are run by opening `/tests` folder in a browser, CSP is only applied if `delivery` option contains `"meta"` for the build environment.

It's obvious that this ends in a bad developer experience. E.g. the tests might pass locally but fail on CI.

It should be guaranteed that tests are always using the CSP for test environment regardless how they are executed. This requires two changes:

1. CSP meta tag must be set for `tests/index.html` regardless of the environment. This is currently only ensured for tests environment: https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/lib/utils.js#L114-L120 One solution might be to return CSP meta tag for `tests-head` in `contentFor` hook if not already returned for `head`.
2. Express server must always use CSP for test environment for `tests/` path.

jelhan:
This ensures that CSP configured for test environment is always used regardless how the tests are executed (`ember test`, `ember test --server`, `ember serve` and opening `http://localhost:4200/tests` in a browser). It also removes some CSP only required for tests from the CSP meta tag in `index.html` if build includes tests and from CSP header used for serving `index.html` with `ember serve`.

Closes #119

### DIFF
--- a/index.js
+++ b/index.js
@@ -175,7 +175,12 @@ module.exports = {
     }
 
     // inject CSP meta tag
-    if (type === 'head' && this._config.delivery.indexOf('meta') !== -1) {
+    if (
+      // if addon is configured to deliver CSP by meta tag
+      ( type === 'head' && this._config.delivery.indexOf('meta') !== -1 ) ||
+      // ensure it's injected in tests/index.html to ensure consistent test results
+      type === 'test-head'
+    ) {
       this.ui.writeWarnLine(
         'Content Security Policy does not support report only mode if delivered via meta element. ' +
         "Either set `ENV['ember-cli-content-security-policy'].reportOnly` to `false` or remove `'meta'` " +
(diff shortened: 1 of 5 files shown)


rwjblue:
Awesome, thank you!

PR #125, opened 2019-12-30
Fix: Throws if development server is used with build that does not contain tests

### LINKED ISSUE #124: Cannot read property 'policy' of undefined
This happens because of this new addition
`https://github.com/rwjblue/ember-cli-content-security-policy/blame/master/index.js#L133`

jelhan:
Fixes #124

### DIFF
--- a/index.js
+++ b/index.js
@@ -175,9 +175,9 @@ module.exports = {
     }
 
     app.use((req, res, next) => {
-      // Use policy for test environment if
-      // - the request is for tests and
-      // - the build include tests
+      // Use policy for test environment if both of these conditions are met:
+      // 1. the request is for tests and
+      // 2. the build include tests
       let isRequestForTests = req.originalUrl.startsWith('/tests') && app.tests;
       let config = isRequestForTests ? this._configForTest : this._config;
       let policyString = isRequestForTests ? this._policyStringForTest : this._policyString;
--- a/index.js
+++ b/index.js
@@ -130,7 +130,11 @@ module.exports = {
     // support live reload, executing tests in development enviroment via
     // `http://localhost:4200/tests` and reporting CSP violations on CLI.
     let policyObject = this._config.policy;
-    let policyObjectForTest = this._configForTest.policy;
+
+    // Policy object for tests is only calculated if build includes tests
+    // (`app.tests === true`). If it hasn't been calculated at all, there
+    // is no need to recalculate it.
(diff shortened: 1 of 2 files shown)


sandstrom:
Would an alternative solution be that people disabled this addon only for their test-environment? 

```js
module.exports = function(environment) {
  if (environment === 'test') {
    contentSecurityPolicy.enabled = false;
    // …

**Comment by jelhan** [comment]
> Would an alternative solution be that people disabled this addon only for their test-environment?

That's possible but won't solve the issue. Actually we want to support having the addon enabled for tests and still be able to serve builds not including tests:

- The addon must be enabled for test environment in order to get CSP validation checks as part of your test suite.
- `ember serve -prod` is a common scenario in which a build without tests is served by built-in web server.

File: index.js [hunk 1]
```diff
@@ -154,17 +160,25 @@ module.exports = {
...
+
+    if (policyObjectForTest) {
+      this._policyStringForTest = buildPolicyString(policyObjectForTest);
+    }
 
     app.use((req, res, next) => {
-      let isRequestForTests = req.originalUrl.startsWith('/tests');
+      // Use policy for test environment if
+      // - the request is for tests and
+      // - the build include tests
```
sandstrom: Minor, but could clarify:

```
      // Use policy for test environment if both of these conditions are met:
      // 1. the request is for tests and
      // 2. the build include tests
```

node-tests/e2e/test-support-test.js
```diff
@@ -185,4 +185,28 @@ describe('e2e: provides test support', function() {
 
     expect(code).to.equal(0);
   });
+
+  it('does not break development server for builds not including tests', async function() {
```
sandstrom:
```suggestion
 // One common scenario is when running the server in production mode locally, i.e. 
 // you are doing the production build on your local machine, connecting to your actual production server,
 // for example via `ember serve -prod`. In these cases we don't want this addon to break.
  it('does not break development server for builds not including tests', async function() {
```

index.js
jelhan: Thanks a lot for your suggestion. Applied it in 952f3fccb4b9bf4b2d8a80cfd6b9991a25435eec.

node-tests/e2e/test-support-test.js
jelhan: Applied that suggestion on 952f3fccb4b9bf4b2d8a80cfd6b9991a25435eec as well.

PR #120, opened 2019-12-10
WIP: upgrade dependencies

jelhan:
Upgrades all dependencies to their latest versions.

PR #126, opened 2020-01-07
upgrade dependencies

jelhan:
This upgrades all dependencies to their latest versions and applying all changes in the blueprints that don't break the support policy (Ember 2.18+). If I recall correctly only the changes to native classes in dummy app are not taken from new blueprints cause that would require [ember-native-class-polyfill](https://github.com/pzuraq/ember-native-class-polyfill) only for testing.

Have done all the upgrades step by step after the hassle in #120. Feeling like RenovateBot. :laughing: I have no idea why it has been that smooth this time to be honest.


jelhan:
@sandstrom Do you may have time for a review? I prefer to have another pair of eyes on my pull requests rather than merging them on my own.

.npmignore
```diff
@@ -12,6 +12,7 @@
 /.env*
 /.eslintignore
 /.eslintrc.js
+/.git/
```
sandstrom: this is ignored by default, no need to add:
https://docs.npmjs.com/using-npm/developers.html#keeping-files-out-of-your-package

.template-lintrc.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
 
 module.exports = {
-  extends: 'recommended'
+  extends: 'octane'
```
sandstrom: Don't know the implication of this, e.g. I don't know if this will impact consuming applications, but probably not.

.travis.yml
```diff
@@ -3,27 +3,22 @@ language: node_js
...
   global:
     # See https://git.io/vdao3 for details.
     - JOBS=1
 
-branches:
-  only:
-    - master
-    # npm version tags
-    - /^v\d+\.\d+\.\d+/
-
```
sandstrom: Is there a reason why we're removing this?

config/ember-try.js
```diff
@@ -2,83 +2,108 @@
...
+      // `ember try:each` manually or from a customized CI config will run it
+      // along with all the other scenarios.
+      {
+        name: 'ember-default',
+        npm: {
+          devDependencies: {}
+        }
+      },
+      {
+        name: 'ember-default-with-jquery',
```
sandstrom: Do we need to test with jQuery? We're not using jQuery anywhere in this addon afaik

.template-lintrc.js
jelhan: This should only affect template linting for this addon.

.npmignore
jelhan: It was added to the addon blueprint in 3.13: https://github.com/ember-cli/ember-addon-output/commit/7aea9e08d5711cb79a8eb86ef0281f9de63b621f#diff-0fd4ef892d9d4990033701887c2f9bcc

It was implemented in this commit: https://github.com/ember-cli/ember-cli/commit/14ba465649362a7f5903fad958de97a520ea900e The commit message says that at least in one case the `/.git/` directory was not ignored by default. But it seemed to be unclear why.

.travis.yml
jelhan: Ups. This was only required for testing my feature branch. Restored in 21a229c6c996bbac16a3a1eb247953afb251188b.

config/ember-try.js
jelhan: It's part of the blueprint for addons: https://github.com/ember-cli/ember-addon-output/blob/master/config/ember-try.js But I agree that it shouldn't be needed for this addon. I don't care if we remove it or not.

jelhan:
@sandstrom Thanks a lot for your review. I hope I addressed all points. Let me know if you have follow up questions / concerns / change requests. If not, please feel free to _squash and merge_.

PR #128, opened 2020-01-27
Don't add nonce to script-src when it already contains 'unsafe-inline'

### LINKED ISSUE #127: Problem with using 'unsafe-inline' in script-src
We get consistent CSP reports while running the development server. From what I've been able to tell, this is because in https://github.com/rwjblue/ember-cli-content-security-policy/pull/122, a nonce value was unconditionally added to `srcript-src` in all apps with tests.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src), as soon as you set a nonce, `'unsafe-inline'` is ignored. We were relying on `'unsafe-inline'`, since it's not trivial to get the hash of all inline scripts in our app at build time.

I have to admit I do not understand exactly what problem https://github.com/rwjblue/ember-cli-content-security-policy/pull/122 was trying to solve. For us, it would be a good solution to omit the nonce if the `srcipt-src` directive contains `'unsafe-inline'`.

If you like that solution, I can make a PR for it.

joukevandermaas:
Fixes #127.

I added a test for both the positive (should add nonce) and negative (should not add nonce) case, because I could not find a test for the former and I wanted to make sure I didn't accidentally break that scenario.

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,8 +83,13 @@ module.exports = {
         let runConfigForTest = project.config('test');
         let configForTest = calculateConfig('test', ownConfigForTest, runConfigForTest, ui);
 
-        // add static nonce required for tests
-        appendSourceList(configForTest.policy, 'script-src', `'nonce-${STATIC_TEST_NONCE}'`);
+        // add static nonce required for tests, but only if if script-src
+        // does not contain 'unsafe-inline'. if a nonce is present, browsers
+        // ignore the 'unsafe-inline' directive.
+        let scriptSrc = configForTest.policy['script-src'];
+        if (!(scriptSrc && scriptSrc.includes("'unsafe-inline'"))) {
+          appendSourceList(configForTest.policy, 'script-src', `'nonce-${STATIC_TEST_NONCE}'`);
+        }
 
         // testem requires frame-src to run
         configForTest.policy['frame-src'] = ["'self'"];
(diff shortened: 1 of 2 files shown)


jelhan:
Closing and reopening so that Travis will hopefully pick this one up.

jelhan:
TravisCI refuses to run this pull request with an `Abuse detected` error:

abuse-travis

Haven't seen this before. A quick research on Google said that it might be fixed if you login to https://travis-ci.com/ once with your GitHub account. You could restart TravisCI by closing and reopening this pull request afterwards. Could you please try it?

joukevandermaas:
I logged in and now I see this notification:

image

I wonder what I did to them :smile: I will contact their support.

joukevandermaas:
I managed to clear up the issue with travis support, and it seems they unblocked my email. :tada:

jelhan:
@joukevandermaas Thanks a lot for the fix. Not sure if we should backport it to `v1`. If not we should at least make sure to have a first release candidate of `v2` soon. @rwjblue any preference? v2 is ready for first release candidate from my side. Just need to get the two pending pull requests merged before.

urbany:
Just noticed this bug has been hitting me for a few months, is it safe to update to v2? or can we have a backport please? Thank you very much in advance!

jelhan:
> Just noticed this bug has been hitting me for a few months, is it safe to update to v2? or can we have a backport please? Thank you very much in advance!

Sorry for not replying back earlier. Upgrading to v2 should be safe. I think it's more stable than v1 at this point of time even though still considered a pre-release.

PR #134, opened 2020-03-19
Prevent unnecessary meta + reportOnly warning

reidab:
This warning was being output on dev server startup when `test-head` was rendered, even if both of the settings mentioned in the message were set to avoid it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -236,8 +236,8 @@ module.exports = {
       if (this._config.reportOnly && this._config.delivery.indexOf('meta') !== -1) {
         this.ui.writeWarnLine(
           'Content Security Policy does not support report only mode if delivered via meta element. ' +
-          "Either set `ENV['ember-cli-content-security-policy'].reportOnly` to `false` or remove `'meta'` " +
-          "from `ENV['ember-cli-content-security-policy'].delivery`.",
+          "Either set `reportOnly` to `false` or remove `'meta' from `delivery` in " +
+          '`config/content-security-policy.js`.',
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -233,12 +233,14 @@ module.exports = {
...
-        config.reportOnly
-      );
+      if (this._config.reportOnly && this._config.delivery.indexOf('meta') !== -1) {
+        this.ui.writeWarnLine(
+          'Content Security Policy does not support report only mode if delivered via meta element. ' +
+          "Either set `ENV['ember-cli-content-security-policy'].reportOnly` to `false` or remove `'meta'` " +
+          "from `ENV['ember-cli-content-security-policy'].delivery`.",
+          config.reportOnly
+        );
+      }
```
jelhan: Ups. I fear I used the `writeWarnLine` of [`console-ui`](https://github.com/ember-cli/console-ui) as if it is `warn` from `@ember/debug`. :grimacing: `writeWarnLine` doesn't seem to have a second argument, that controls if the warning is shown or not.

Also the message is outdated. It still references the deprecated configuration in `config/environment.js`. It should refer to `config/content-security-policy.js`. Seems like I missed to update it.

Do you have time to fix the message and remove the second argument?

jelhan:
@reidab Updated the warning text in 44edd971334ea20f1526cf18ec37b1ca6b4514b6. Hope it was okay to push to your pull request.

reidab:
@jelhan that's great! thanks for merging!

Issue #133, opened 2020-03-19
Dev server CSP items are missing when requested with Accept: text/html using Fastboot on master


reidab:
Since it looks like a beta of 2.x is coming soon, I was testing out the `master` branch (https://github.com/rwjblue/ember-cli-content-security-policy/commit/5dc741e3e7739899c5ee97eea5e86450becb1ede) to see if it would solve something similar to #127 that I was running into. It did, but I'm now running into some bizarre behavior around the `localhost` CSP injection for the dev server with Fastboot.

When requested with an `Accept` header specifying `*/*` (the default from `curl`), it looks like the raw `index.html` output is returned along with the expected CSP headers. When requesting with an `Accept` header that includes `text/html` (like a browser would use) the Fastboot-rendered body is returned but the `content-security-policy` header no longer includes `localhost:4200 0.0.0.0:4200` and other directives that were injected for local dev server usage. Strangely, the `x-content-security-policy` header still includes these values.

The calls to `setHeader` in ember-cli-content-security-policy certainly seem to be called with the appropriate values, so maybe something's rewriting the headers further downstream? I'd say it was failing to detect `liveReload`, but that wouldn't explain `x-content-security-policy` sneaking through unscathed…

Anyhow, I'm not sure if this issue is with ember-cli-content-security-policy or ember-cli-fastboot, but it was baffling enough that I wanted to write it up. 😁 

## Expected

Both `content-security-policy` and `x-content-security-policy` to be the same CSP string including `script-src` and `connect-src` entries for localhost.

## Actual

`x-content-security-policy` is delivered with the localhost entries included

```
x-content-security-policy: default-src 'none'; script-src 'self' localhost:4200
0.0.0.0:4200; font-src 'self'; connect-src 'self' ws://localhost:4200
ws://0.0.0.0:4200; img-src 'self'; style-src 'self'; media-src 'self';
```

`content-security-policy` is delivered without the localhost entries

```
content-security-policy: default-src 'none'; script-src 'self'; font-src 'self'; 
connect-src 'self'; img-src 'self'; style-src 'self'; media-src 'self';
```

## Repro / Investigation

```
$ ember new csp-test
installing app
Ember CLI v3.17.0
[…]

$ cd csp-test
$ ember install ember-cli-content-security-policy@1.1.1
$ ember s
```

At this point, I'm running 1.1.1 without Fastboot. CSP headers are returned as expected regardless of the `Accept` header.

<details>
[...]

PR #135, opened 2020-03-23
do not set X-Content-Security-Policy header

jelhan:
All major browsers support Content-Security-Policy header since a long time. Only Internet Explorer 11 does not support it. But it also only supports `sandbox` directive by X-Content-Security-Policy header. The sandbox is not useful at all for Ember applications cause it could only be used to disable all scripts. As an Ember applicaiton is not usable at all without JavaScript that is not a reasonable configuration.

This resolves the inconsistency between middleware used in Ember CLI's development server and FastBoot initalizier described in #133.


sandstrom:
@jelhan Great!

PR #144, opened 2020-04-14
rename express app variable to expressApp

jelhan:
Having both the `EmberApp` instance and the express server named `app` is causing confusion and leading to bugs. Renamed the variable used for express server to `expressApp`.


rwjblue:
:thumbsup:

PR #154, opened 2020-11-04
refactor tests to use Ember Addon Tests

jelhan:
This merge requests refactors the test suite to use [Ember Addon Tests](https://github.com/jelhan/ember-addon-tests) instead of [Ember CLI Addon Tests](https://github.com/tomdale/ember-cli-addon-tests).

Ember CLI Addon Tests were unstable and slow. CI pipeline was breaking due to small changes in timing and order of tests. Some tests needed to be executed in different processes to prevent leakage. This caused the tests to be sometimes grouped to not break Ember CLI Addon Tests rather than for readability.

Ember Addon Tests could be seen as a modernized rewrite of Ember CLI Addon Tests utilizing yarn workspaces. It should also provide a better performance but I haven't benchmarked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,7 @@ node_js:
   # so that your addon works for all apps
   - "10"
 
-sudo: false
-dist: trusty
+dist: xenial
 
 addons:
   chrome: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ jobs:
       env: EMBER_CLI_VERSION=~3.4.0
     - env: EMBER_CLI_VERSION=~3.8.0
     - env: EMBER_CLI_VERSION=~3.12.0
+    - env: EMBER_CLI_VERSION=~3.16.0
+    - env: EMBER_CLI_VERSION=~3.20.0
     - env: EMBER_CLI_VERSION=latest
     - env: EMBER_CLI_VERSION=beta
 
--- a/index.js
+++ b/index.js
@@ -105,7 +105,7 @@ module.exports = {
     // - application has ember-cli-fastboot dependency.
     this._needsFastBootSupport = this._config.enabled &&
       this._config.delivery.includes('header') &&
-      new VersionChecker(this.project).for('ember-cli-fastboot').exists();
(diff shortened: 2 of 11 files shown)


index.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
     // - application has ember-cli-fastboot dependency.
     this._needsFastBootSupport = this._config.enabled &&
       this._config.delivery.includes('header') &&
-      new VersionChecker(this.project).for('ember-cli-fastboot').exists();
+      this.project.findAddonByName('ember-cli-fastboot') !== null;
```
jelhan, line 108: Ember CLI Version Checker does not work well within yarn workspaces. It reports wrongly that a dependency exists if that dependency is installed in the workspace even if it's not a direct nor indirect dependency of the package.

jelhan:
@sandstrom This is the pull request that I promised in https://github.com/rwjblue/ember-cli-content-security-policy/issues/100#issuecomment-721736159.

sandstrom:
@jelhan Awesome! 💯 🎉 

I've looked at the diff and looks great! But I don't have much experience with any of the two test frameworks, so I can't really give any useful feedback or extra pair of eyes 👀. All I could hope to find would be spelling mistakes etc.

.travis.yml
```diff
@@ -5,8 +5,7 @@ node_js:
   # so that your addon works for all apps
   - "10"
 
-sudo: false
-dist: trusty
+dist: xenial
```
jelhan, line 8: CI failed due to outdated yarn version. This change is inline with latest addon blueprint.

jelhan:
It seems as if one test is still flickering even after using a more recent yarn version:

> test support warns if dependencies are not supported warns if QUnit version is to old

Also the tests for Ember CLI beta are failing. But for that version not randomly but constantly.

I can't reproduce either of these issues locally. To be honest I hope they are related to TravisCI.

But even the current situation is better than the one before. Ember CLI addon tests caused the test suite to break apart on nearly every change. This slowed down development dramatically.

If no concerns are raised I will merge this one by the end of today.

PR #152, opened 2020-09-08
append frame-src config in test mode

chbonser:
Addresses https://github.com/rwjblue/ember-cli-content-security-policy/issues/151

Note this works in my case because my app has `frame-src` defined. From the failing tests it appears `appendSourceList` does not handle the case of a missing `frame-src` properly. Should `appendSourceList` get smarter or should this problem get solved via another approach?

PR #156, opened 2020-11-09
Remove existing 'none' keyword when applying to source list

jelhan:
The bug was uncovered in #152 but also affects other parts as shown in the added test.


jelhan:
Ember CLI beta scenario is failing. This was already the case in #154, which refactored the testing suite. So it doesn't seem to be  related to this change.

Issue #148, opened 2020-07-16
Should remove `report-uri` from CSP if used in meta tag


jelhan:
The `report-uri` directive is only supported if CSP is delivered using HTTP header. It's not allowed if CSP is delivered using meta tag. If included in a CSP meta tag the browser throws an error. If the user configures the addon to deliver CSP both with HTTP header and meta tag he doesn't have any chance to prevent that error.

To avoid confusion we should strip the `report-uri` directive from CSP if included in meta tag. We should instead add a build time warning if user explicitly sets the `report-uri` directive but `delivery` option does not include `'header'`.

rwjblue:
I think stripping when you are using _both_ delivery mechanism is probably fine, but I have a couple questions:

* why warn vs error when they are only using `header` delivery but have `report-uri`? 
* why would you use both delivery mechanisms?

jelhan:
>     * why warn vs error when they are only using `header` delivery but have `report-uri`?

There shouldn't be a warning if `delivery` includes `'header'` as having a `report-uri` directive is valid in that case. I guess you meant the opposite? In that case a warning seems to be more appropriate as the error would not affect build output. The `report-uri` directive would be stripped anyways from CSP in header. Only reason to warn at all is that developers may get confused by such a configuration.

>     * why would you use both delivery mechanisms?

Good question. I'm not aware of any actual use case. But it's currently supported by our configuration interface.

rwjblue:
> I guess you meant the opposite?

D'oh, sorry, yes.

rwjblue:
OK, I'm sold. 😸 

Warning seems fine (definitely better than what we do right now).

sandstrom:
Good idea!

You probably know already, but `report-uri` is deprecated in favour of `report-to`[1].

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri

jelhan:
This was partly addressed by #158. Before the addon was _adding_ a `report-uri` directive even for CSP shipped by meta tag. After this change `report-uri` to report CSP violations in development server is only added to CSP shipped via HTTP header.

A `report-uri` directive included in CSP defined by the user is still not removed from CSP delivered via meta tag. But it should as a user does not have the possibility to define different CSP for delivery via meta and HTTP header. Additionally the CSP is applied as a meta tag in tests even if user hasn't configured the addon to deliver CSP via meta tag.

[...]

PR #158, opened 2020-11-15
Support live reload and add optional debug log

### LINKED ISSUE #146: Development server must be restarted after each configuration change
Changes to the configuration in `config/ember-cli-content-security-policy.js` are not reflected by a running development server instance even so they trigger a rebuild. A restart of the development server is required. This should be either documented or resolved.

jelhan:
Ember CLI Content Security Policy had cached its configuration. This broke live reload because the cache was not invalidated when the addon's configuration file changed. This was reported in #146. Beside causing this bug the caching also added a lot of complexity.

This pull request removes the caching all together. Instead the configuration is calculated whenever it's needed. This has a negative impact on performance. Especially because configuration file is read multiple times - even for the same build. But I don't think the performance impact is noticeable by end-user because the file is still read less than 10 times, it's very small and filesystem caching should speed up subsequent reads.

It also avoid adding `report-uri` to a policy, which is delivered via meta tag. This addresses #148 at least partly.

Additionally it introduces a debug output using [debug](https://www.npmjs.com/package/debug) NPM package. User can opt-in to debug output by setting `DEBUG` environment variable to `ember-cli-content-security-policy`, e.g. `DEBUG=ember-cli-content-security-policy ember build`.

The debug package is used in [ember-auto-import](https://github.com/ef4/ember-auto-import/blob/7610021b3c592bcaa5a649a1d6fa799826f857d3/packages/ember-auto-import/package.json#L42), [embroider](https://github.com/embroider-build/embroider/blob/25d6042c69cbce8355b1a9fa0114447425692e8d/packages/core/package.json#L37) and [ember-data](https://github.com/emberjs/data/blob/4fc1b0774887bb7365113cc655212e10a14ca820/package.json#L71). It seems to be a well established tool in Ember ecosystem.

Closes #146

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,8 @@ const {
   appendSourceList,
   buildPolicyString,
   calculateConfig,
-  isIndexHtmlForTesting,
+  debug,
+  getEnvironmentFromRuntimeConfig,
   readConfig,
 } = require('./lib/utils');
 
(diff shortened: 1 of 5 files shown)


node-tests/e2e/test-support-test.js
```diff
@@ -62,34 +62,90 @@ describe('e2e: provides test support', function() {
     );
   });
 
-  afterEach(async function() {
-    await removeConfig(testProject);
-  });
```
jelhan, line 67: Global `afterEach` caused issues because it was running after *every* test even if part of a `describe` hook. Some `describe` blocks are running multiple tests against the same running development server. After fixing live reload support this showed strange timing issues. Moved all tests into describe hooks and avoided global `afterEach` to have a more fine granular control.

node-tests/e2e/test-support-test.js
```diff
@@ -62,34 +62,90 @@ describe('e2e: provides test support', function() {
...
+  describe('causes tests to fail on CSP violations', async function() {
+    afterEach(async function() {
+      await removeConfig(testProject);
+    });
```
jelhan, line 68: Global `afterEach` caused issues because it was running after *every* test even if part of a `describe` hook. Some `describe` blocks are running multiple tests against the same running development server. After fixing live reload support this showed strange timing issues. Moved all tests into describe hooks and avoided global `afterEach` to have a more fine granular control.

node-tests/e2e/test-support-test.js
```diff
@@ -62,34 +62,90 @@ describe('e2e: provides test support', function() {
...
+        headers: {
+          'Accept': 'text/html'
+        }
+      });
+
+      expect(response.statusCode).to.equal(200);
+      expect(responseForTests.statusCode).to.equal(200);
+
+      await testProject.stopEmberServer();
+    });
```
jelhan, line 129: These tests are just moved from another position within the file. Beside that they are untouched if I recall correctly.

index.js
```diff
@@ -299,19 +285,64 @@ module.exports = {
...
+  _liveReloadConfiguration: null,
+
+  // returns the config for a given environment and delivery method
+  _getConfigFor(environment) {
+    debug(`Calculate configuration for environment ${environment}`);
+
+    const { project } = this;
+    const { ui } = project;
+    const ownConfig = readConfig(project, environment);
+    const runConfig = this._runConfig;
```
jelhan, line 311: Haven't researched if the run-time configuration is available through some public API on the `project` or `app`. I'm not a big fan of storing it in `config` hook and using it later. But I didn't wanted to touch to many parts at the same time.

sandstrom (commented):
Looks good to me!

jelhan:
Pulled out the changes of the tests in #163 and #164, resolved merge conflicts caused by adding Prettier in #162 and rebased against latest master. I hope that tests are passing due to #161. Let's see.

Issue #157, opened 2020-11-14
Setup Prettier


jelhan:
Prettier should be used. Setup should follow the description given in [RFC 628: Prettier](https://emberjs.github.io/rfcs/0628-prettier.html).

zg3d:
Can I help with this issue.

jelhan:
@zg3d That would be great! If it's okay for you, it may be helpful to wait until #158 is merged. It touches many places and I would fear a huge amount of merge conflicts if we do both in parallel. I hope that I can merge it soon.

zg3d:
Alright! I will wait thanks!

jelhan:
@zg3d Seems I can not land #158 as soon as I hoped. Uncovered some other issues with test suite and CI pipeline when running against other Ember CLI versions. Feel free to provide the Prettier merge request as soon as it fits for you. Will rebase my feature branch afterwards if needed.

jelhan:
Closed by #162

PR #162, opened 2020-11-18
Setup Prettier

zg3d:
Fixes issue #157 

-added .pretterrc.json
-added prettier script in package.json
-the script was ran on the repo
- added formatting command in contributing document

Please let me know if I need to make changes


package.json
```diff
@@ -19,7 +19,8 @@
     "start": "ember serve",
     "test": "ember test",
     "test:all": "ember try:each",
-    "test:node": "mocha node-tests/**/*-test.js"
+    "test:node": "mocha node-tests/**/*-test.js",
+    "format": "prettier --write \"./**/*.{md,json,js,yml}\""
```
jelhan, line 23: I would prefer to setup Prettier as ESLint plugin as described in the RFC: https://emberjs.github.io/rfcs/0628-prettier.html#detailed-design My main motivation is that we don't need to change much when this RFC is implemented in Ember CLI.

.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "printWidth": 120,
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "semi": true,
+  "tabWidth": 2,
+  "bracketSpacing": false
+}
```
jelhan, line 8: Any reasons for this specific configuration? I would prefer to use the configuration as defined by Prettier RFC: https://emberjs.github.io/rfcs/0628-prettier.html#prettier Mostly to use the same as other Ember addons.

jelhan (changes requested):
Thanks a lot for picking this one up! :clap:

Would it okay for you to align it more with Prettier RFC? I think this would be more future proof and cause less noise when Prettier is added to default blueprints.

zg3d:
I'll implement all the changes

jelhan (approved):
Thanks a lot!

PR #165, opened 2020-11-23
upgrade dependencies

jelhan:
Upgrades all dependencies to latest versions.


jelhan:
Let's not waste time doing something manual which could be automated: #173

PR #195, opened 2021-03-15
Fix typo `form-ancestors` -> `frame-ancestors` in readme

nicomihalich:
I believe `form-ancestors` should actually be `frame-ancestors` per: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#navigation_directives


jelhan (approved):
Good catch! Thanks a lot for fixing asap.

PR #197, opened 2021-04-15
Support Ember CLI >= 3.25.1

snewcomer:
ref https://github.com/ember-cli/ember-cli/pull/9467/files

https://regex101.com/r/1P9NRv/1

PR #198, opened 2021-04-15
[Deps]: ember-cli 3.26.0

snewcomer:


CONTRIBUTING.md
```diff
@@ -2,29 +2,24 @@
 
 ## Installation
 
-- `git clone https://github.com/rwjblue/ember-cli-content-security-policy.git`
-- `cd ember-cli-content-security-policy`
-- `yarn install`
+* `git clone <repository-url>`
```
jelhan, line 5:
```suggestion
* `git clone https://github.com/rwjblue/ember-cli-content-security-policy.git`
```

config/ember-try.js
```diff
@@ -1,46 +1,25 @@
...
 
 const getChannelURL = require('ember-source-channel-url');
+const { embroiderSafe, embroiderOptimized } = require('@embroider/test-setup');
 
 module.exports = async function () {
   return {
     useYarn: true,
     scenarios: [
       {
-        name: 'ember-lts-2.18',
```
jelhan, line 10: The new scenarios for Ember 3.16, 3.20 and Embroider should be added without replacing the existing scenarios. We need them to ensure that we do not break compatibility commitment without noticing. Maybe we should add a Ember 3.12 scenario as well to cover all LTS versions since 2.18.

index.js
```diff
@@ -296,12 +296,23 @@ module.exports = {
...
+          result = /<script>\s*Ember.assert\(.*EmberENV.TESTS_FILE_LOADED\);\s*<\/script>/.test(
+            entry
+          );
+        } else {
+          result = /<script>.*?Ember\.assert\(.*EmberENV\.TESTS_FILE_LOADED\);\s*}\);<\/script>/.test(
             entry
-          )
-        ) {
+          );
+        }
```
jelhan, line 313: Do I get it right that this is adding compatibility with Ember CLI >= 3.25.1 and therefore fixes floating dependencies test scenario?

jelhan (changes requested):
Thanks a lot for working on this! Just two small changes.

index.js
jelhan: Just reviewed #197, which is about this change only. I guess you just included that one here to ease testing. I have just merged it.

config/ember-try.js
jelhan: I fear I got confused. The addon does not seem to run Ember Try at all. It uses a custom setup to test against multiple Ember CLI versions: https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/.github/workflows/ci.yml#L145

This implicitly covers ember-source versions as well. But does not include ember-source@2.18.

It also means that Embroider tests are not run yet. Even after merging this one.

I guess we need to revisit the compatibility guarantee and test suite. But it could be done independently from this merge request.

Issue #200, opened 2021-04-21
Inconsistent test results


jelhan:
The following test is expected to be broken for Ember CLI 3.26.1 due to #199 

https://github.com/rwjblue/ember-cli-content-security-policy/blob/93924fa54dd472de89ea0b0f5d1f126c40758c9a/node-tests/e2e/test-support-test.js#L69-L75

It is in-deed failing locally when executed using global `mocha`:

```
$ DEBUG=ember-addon-tests mocha node-tests/**/*-test.js --grep "tests are passing for untouched application"


  e2e: provides test support
  ember-addon-tests initalize workspace +0ms
  ember-addon-tests created workspace directory at /tmp/ember-addon-tests-ZNw1lm +1ms
  ember-addon-tests configured workspace +1ms
  ember-addon-tests Indentifying packages under test for package located at /home/main/Code/ember-ecosystem/ember-cli-content-security-policy +0ms
  ember-addon-tests Project does not use Yarn workspaces +194ms
  ember-addon-tests link package ember-cli-content-security-policy located at /home/main/Code/ember-ecosystem/ember-cli-content-security-policy into yarn workspace +1ms
  ember-addon-tests linked all packages under test into yarn workspace +26ms
  ember-addon-tests install dependencies of packages under test +0ms
  ember-addon-tests installed dependencies of packages under test +38s
  ember-addon-tests initalized workspace at /tmp/ember-addon-tests-ZNw1lm +0ms
  ember-addon-tests created test project at /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm +160ms
  ember-addon-tests execute command `ember new a8nOQZm --skip-git --yarn --directory /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm` in /home/main +3ms
  ember-addon-tests executed command `ember new a8nOQZm --skip-git --yarn --directory /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm` in /home/main +23s
  ember-addon-tests execute command `yarn install` in /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm +3ms
  ember-addon-tests executed command `yarn install` in /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm +5s
  ember-addon-tests execute command `yarn remove ember-auto-import` in /tmp/ember-addon-tests-ZNw1lm/test-projects/a8nOQZm +0ms
[...]

PR #201, opened 2021-04-23
Support Ember CLI >= 3.26.0 and match injected script element by all supported Ember CLI versions with same RegExp

### LINKED ISSUE #199: Tests fail after installing this addon because of `ember-cli`
`ember-cli` adds a script tag (see [here](https://github.com/ember-cli/ember-cli/blob/v3.26.1/lib/utilities/ember-app-utils.js#L132)) when running tests. This fails the tests unless one adds `unsafe-inline` to `script-src`. What are we supposed to do about that? I guess that should be fixed in `ember-cli`?

snewcomer:
close #199

### DIFF
--- a/index.js
+++ b/index.js
@@ -296,21 +296,10 @@ module.exports = {
 
     // Add nonce to <script> tag inserted by Ember CLI to assert that test file was loaded.
     if (type === 'test-body-footer') {
-      let emberCliDependency = new VersionChecker(this.project).for(
-        'ember-cli'
-      );
-
       existingContent.forEach((entry, index) => {
-        let result;
-        if (emberCliDependency.exists() && emberCliDependency.lt('3.25.1')) {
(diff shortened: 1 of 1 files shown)


jelhan (approved):
Thanks a lot for fixing this.

Please excuse that it has taken that long for me to reply. As documented in #200 I lost trust in the CI pipeline. I wanted to get that fixed again at least somehow to verify that the new regular expression is working for all supported Ember CLI versions.

#202 fixes the CI setup. I merged this one it. The CI is still unstable for Ember CLI 3.8. Therefore it is not merged yet. But it has proven that this fix is correct. :tada:

jelhan:
Released as [v2.0.0-4](https://github.com/rwjblue/ember-cli-content-security-policy/releases/tag/v2.0.0-4)

PR #271, opened 2021-12-02
Fastboot instance initializer throws if reportOnly config is false

JoeyBG:
The current behaviour makes it impossible to disable `reportOnly` mode, we actually want to ensure the configuration is properly defined, not that it's true.

### DIFF
--- a/fastboot/instance-initializers/content-security-policy.js
+++ b/fastboot/instance-initializers/content-security-policy.js
@@ -9,7 +9,9 @@ function readAddonConfig(appInstance) {
   //       if already available through CSP meta element
   assert(
     'Required configuration is available at run-time',
-    addonConfig && addonConfig['reportOnly'] && addonConfig['policy']
+    addonConfig &&
+      Object.prototype.hasOwnProperty.call(addonConfig, 'reportOnly') &&
+      Object.prototype.hasOwnProperty.call(addonConfig, 'policy')
   );
 
   return config['ember-cli-content-security-policy'];
--- a/fastboot/instance-initializers/content-security-policy.js
+++ b/fastboot/instance-initializers/content-security-policy.js
@@ -11,7 +11,7 @@ function readAddonConfig(appInstance) {
     'Required configuration is available at run-time',
     typeof addonConfig === 'object' &&
       typeof addonConfig.reportOnly === 'boolean' &&
-      typeof addonConfig.policy === 'object'
+      typeof addonConfig.policy === 'string'
   );
 
   return config['ember-cli-content-security-policy'];
--- a/fastboot/instance-initializers/content-security-policy.js
+++ b/fastboot/instance-initializers/content-security-policy.js
@@ -9,9 +9,9 @@ function readAddonConfig(appInstance) {
   //       if already available through CSP meta element
   assert(
     'Required configuration is available at run-time',
-    addonConfig &&
-      Object.prototype.hasOwnProperty.call(addonConfig, 'reportOnly') &&
-      Object.prototype.hasOwnProperty.call(addonConfig, 'policy')
+    typeof addonConfig === 'object' &&
+      typeof addonConfig.reportOnly === 'boolean' &&
+      typeof addonConfig.policy === 'object'
   );
 
   return config['ember-cli-content-security-policy'];
--- a/node-tests/e2e/fastboot-support-test.js
+++ b/node-tests/e2e/fastboot-support-test.js
@@ -104,6 +104,7 @@ describe('e2e: fastboot integration', function () {
         },
       });
 
+      expect(response.statusCode).to.equal(200);
       expect(response.headers).to.include.key(
         'content-security-policy-report-only'
       );
@@ -122,7 +123,7 @@ describe('e2e: fastboot integration', function () {
       await removeConfig(testProject);
     });
 
-    it.only('sets CSP header if served via FastBoot', async function () {
+    it('sets CSP header if served via FastBoot', async function () {
       let response = await request({
         url: 'http://localhost:49742',
         headers: {
@@ -130,6 +131,7 @@ describe('e2e: fastboot integration', function () {
         },
       });
 
+      expect(response.statusCode).to.equal(200);
       expect(response.headers).to.include.key('content-security-policy');
     });
   });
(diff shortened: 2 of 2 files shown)


richthedev (approved):

fastboot/instance-initializers/content-security-policy.js
```diff
@@ -9,7 +9,9 @@ function readAddonConfig(appInstance) {
   //       if already available through CSP meta element
   assert(
     'Required configuration is available at run-time',
-    addonConfig && addonConfig['reportOnly'] && addonConfig['policy']
+    addonConfig &&
+      Object.prototype.hasOwnProperty.call(addonConfig, 'reportOnly') &&
+      Object.prototype.hasOwnProperty.call(addonConfig, 'policy')
```
jelhan, line 14: Wondering if we could assert against the type instead?

```suggestion
    typeof addonConfig === 'object' &&
      typeof addonConfig.reportOnly === 'boolean' &&
      typeof addonConfig.policy === 'string'
```

This seems to be more readable and even make the assertion more strict.

jelhan (changes requested):
Thanks a lot for working on this. Would be great if you could add a test to prevent a regression.

jelhan:
Please ignore the failing test for _Ember CLI beta_. It is not related to your changes. It should be fixed by #272.

fastboot/instance-initializers/content-security-policy.js
JoeyBG: I can't seem to find an existing test that actually runs goes through that codepath. Any ideas?

I'm not too experienced with Ember, I have no clue where to begin to setup tests deeply integrated with Fastboot like that.

I can apply your suggestion, but I believe policy should be 'object' right?

JoeyBG: I changed the implementation as suggested, I will not have time to add tests for this, sorry. Hopefully, this can still be merged and released, as currently, `v2.0.0` crashes on startup when `reportOnly` is set to false in a dev environment.

We're using our fork for now.

jelhan:
The tests are failing with latest changes. Maybe my suggestion isn't working as expected.

Not sure when I have time to look into it. No one is paying me for maintaining this addon. I'm doing it _after_y regular job.

If you have time to debug that would be great.

For tests please have a look at existing Fastboot related tests in `node-tests` folder.

fastboot/instance-initializers/content-security-policy.js
JoeyBG: > The tests are failing with latest changes. Maybe my suggestion isn't working as expected.
> 
> Not sure when I have time to look into it. No one is paying me for maintaining this addon. I'm doing it _after_y regular job.
> 
> If you have time to debug that would be great.
> 
> For tests please have a look at existing Fastboot related tests in `node-tests` folder.

I added a test. It doesn't actually fail with the previous code, since assert only fails in development mode, but as I said I have no clue how to make the test server behave like a dev environment would.

The failing tests were caused by me using `object` for `policy`, I went back to `string` and it works well now.

JoeyBG: @jelhan Just making sure you've seen this notification. No worries if you didn't have time to get to this yet!

node-tests/e2e/fastboot-support-test.js
```diff
@@ -110,6 +110,30 @@ describe('e2e: fastboot integration', function () {
     });
   });
 
+  describe('scenario: with reportOnly = false', function () {
+    before(async function () {
+      await setConfig(testProject, { reportOnly: false });
+      await testProject.runEmberCommand('build');
```
jelhan, line 116: > I added a test. It doesn't actually fail with the previous code, since assert only fails in development mode, but as I said I have no clue how to make the test server behave like a dev environment would.

You can pass flags like build environment to the command:

```suggestion
      await testProject.runEmberCommand('build', '--environment', 'production');
```

But `development` should be the default environment: https://cli.emberjs.com/release/advanced-use/cli-commands-reference/#emberbuild

jelhan (commented):
Thanks for pushing this one forward and adding a test. I'm surprised that the test does _not_ fail without the fix to be honest. Haven't had the chance to try and debug myself yet. 😟

node-tests/e2e/fastboot-support-test.js
```diff
@@ -110,6 +110,30 @@ describe('e2e: fastboot integration', function () {
...
+
+    it.only('sets CSP header if served via FastBoot', async function () {
+      let response = await request({
+        url: 'http://localhost:49742',
+        headers: {
+          Accept: 'text/html',
+        },
+      });
+
+      expect(response.headers).to.include.key('content-security-policy');
```
jelhan, line 133: I had the time to debug why the test is passing on `master` without this change. It is setting the CSP header as expected even though it throws. This could be detected by asserting the HTTP status code as well. It is `500 Internal Server Error` if the assertion is thrown.

```suggestion
      expect(response.statusCode).to.equal(200);
      expect(response.headers).to.include.key('content-security-policy');
```

Would be great if you could add the same assertion before line 107. It is missing there as well.

node-tests/e2e/fastboot-support-test.js
```diff
@@ -110,6 +110,30 @@ describe('e2e: fastboot integration', function () {
...
+      await testProject.runEmberCommand('build');
+      await startServer();
+    });
+
+    after(async function () {
+      await stopServer();
+      await removeConfig(testProject);
+    });
+
+    it.only('sets CSP header if served via FastBoot', async function () {
```
jelhan, line 125: This seems to be a left-over from debugging.

```suggestion
    it('sets CSP header if served via FastBoot', async function () {
```

node-tests/e2e/fastboot-support-test.js
jelhan: Please see my new suggestion about asserting the HTTP status code.

jelhan (changes requested):
I found the time to debug the test. Should be good to merge as soon as you added the missing assertion and removed the left-over from debugging.

JoeyBG:
> I found the time to debug the test. Should be good to merge as soon as you added the missing assertion and removed the left-over from debugging.

@jelhan Done! That makes perfect sense :)

jelhan (approved):
Thanks a lot!

jelhan:
Released this as [v2.0.2](https://github.com/rwjblue/ember-cli-content-security-policy/releases/tag/v2.0.2) just right now. I'm sorry that it took that long. To be honest I just missed that I hadn't released it yet.

PR #274, opened 2021-12-09
fix: pass environment into readConfig

guidojw:
Noticed a small bug while using the csp-headers command: environment was not being passed into the config function.

This PR fixes that!

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -37,7 +37,7 @@ module.exports = {
     run: function (options) {
       let { environment, reportUri } = options;
       let { project, ui } = this;
-      let ownConfig = readConfig(project);
+      let ownConfig = readConfig(project, environment);
       let runConfig = project.config(environment);
       let { reportOnly, policy } = calculateConfig(
         environment,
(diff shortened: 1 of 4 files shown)


jelhan (approved):
Thanks a lot. Looks good to me.

I noticed that API docs for `readConfig` is also missing that argument. May be the reason for which we haven't noticed earlier. If you have time, fixing that one 

If you have time to extend existing test to cover this case, that would be awesome.

If you don't have time for either, that's fine as well.

guidojw:
Added the missing doc and tests! Hope you like what I did in `node-tests/utils`, suggestions are welcome :)

jelhan (approved):
Thanks a lot for adding a test and fixing docs.

> Hope you like what I did in node-tests/utils, suggestions are welcome :)

To be honest I'm a little bit afraid by the added complexity of that special syntax. But I also don't see a better solution. Maybe passing the function as a string. But that also does not help much with readability. Doesn't seem to be an easy one. And your solution good enough for sure. 👍

Not at my computer currently. Will try to release on Monday. If I forget, please ping me.

jelhan:
Released as [v2.0.1](https://github.com/rwjblue/ember-cli-content-security-policy/releases/tag/v2.0.1).

PR #332, opened 2022-08-04
fix: enable serving in separate process from the build

### LINKED ISSUE #331: Breaks if serving separately from build command
Build: `ember build --watch`
Serve: `ember serve --path="dist"`

Error:

```
TypeError: Cannot read property 'tests' of undefined
    at [...]/node_modules/.pnpm/ember-cli-content-security-policy@2.0.3/node_modules/ember-cli-content-security-policy/index.js:127:40
```

The associated line is here: https://github.com/rwjblue/ember-cli-content-security-policy/blob/b6c3e093203acdba8b954442e49c0e91a6d25478/index.js#L127

It appears `this.app` is not available when not also running the build hooks.

runspired:
resolves #331 

If serving from a separate command and your build includes tests you will also want to set the env var `USE_TESTS_CSP=1`.

Optionally if necessary the CSP can be disabled with `DISABLE_CSP` env var.

### DIFF
--- a/index.js
+++ b/index.js
@@ -67,6 +67,9 @@ module.exports = {
   // FastBoot. This one is returned here as default configuration in order to make it
   // available at run time.
   config: function (environment, runConfig) {
+    if (process.env.DISABLE_CSP) {
+      return {};
+    }
     debug('### Cache run-time config locally in config hook');
 
     // store run config to be available later
@@ -101,6 +104,9 @@ module.exports = {
   },
 
   serverMiddleware: function ({ app: expressApp, options }) {
+    if (process.env.DISABLE_CSP) {
+      return;
+    }
     debug('### Register middleware to set CSP headers in development server');
 
     const requiresLiveReload = options.liveReload;
@@ -124,10 +130,13 @@ module.exports = {
       // Use policy for test environment if both of these conditions are met:
       // 1. the request is for tests and
       // 2. the build include tests
-      let buildIncludeTests = this.app.tests;
+      const env = process.env.EMBER_ENV;
+      let buildIncludeTests = this.app
+        ? this.app.tests
+        : process.env.USE_TESTS_CSP || false;
       let isRequestForTests =
         req.originalUrl.startsWith('/tests') && buildIncludeTests;
-      let environment = isRequestForTests ? 'test' : this.app.env;
+      let environment = isRequestForTests ? 'test' : env;
 
       debug(
         buildIncludeTests


runspired:
test failures are unrelated, tests appear to be broken on master

index.js
```diff
@@ -124,10 +130,13 @@ module.exports = {
       // Use policy for test environment if both of these conditions are met:
       // 1. the request is for tests and
       // 2. the build include tests
-      let buildIncludeTests = this.app.tests;
+      const env = process.env.EMBER_ENV;
+      let buildIncludeTests = this.app
```
jelhan, line 134: I'm wondering if we actually need to determine if build includes tests. Maybe it's enough to analyze if the request is for tests? Is there a case of legitim requests for `/tests` if app does not include tests?

index.js
```diff
@@ -101,6 +104,9 @@ module.exports = {
   },
 
   serverMiddleware: function ({ app: expressApp, options }) {
+    if (process.env.DISABLE_CSP) {
```
jelhan, line 107: Any reason to have an environment variable to disable CSP in addon hard-coded? Having it in userland configuration would reduce complexity. A consumer would not need to look into addons documentation but only in project's CSP configuration to discover and understand the environment variable.

jelhan (commented):
Thanks a lot for working on this one. I have a few questions on design.

Didn't had much time to think through it to be honest. Please excuse if the questions sound stupid knowing the details.

index.js
```diff
@@ -67,6 +67,9 @@ module.exports = {
   // FastBoot. This one is returned here as default configuration in order to make it
   // available at run time.
   config: function (environment, runConfig) {
+    if (process.env.DISABLE_CSP) {
```
vfuto, line 70: I had to return after setting `_runConfig`, because https://github.com/rwjblue/ember-cli-content-security-policy/blob/master/lib/utils.js#L96 was null (`Cannot read properties of null (reading 'contentSecurityPolicy')`)
```js
this._runConfig = runConfig;

if (process.env.DISABLE_CSP) {
  return {};
}
```

index.js
runspired: this allows libraries to easily turn it off via config when running the command vs needing to learn about and change config files

index.js
runspired: - you can do a build and then serve tests off the build directly.
- you can name a route tests :(

wozny1989:
@runspired thanks a lot for this fix! 

Any chance of merge it?

PR #365, opened 2024-04-16
fix: Example allows manifest from self

MichalBryxi:
- Given example is quite handy starting copy&paste.
- Unfortunatelly for Chrome with modern Ember apps it gives:

```
assessment:1 Refused to load manifest from 'http://localhost:4200/manifest.webmanifest' because it violates the following Content Security Policy directive: "default-src 'none'". Note that 'manifest-src' was not explicitly set, so 'default-src' is used as a fallback.
```

- This patch makes the README as a nicer starting point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ module.exports = function (environment) {
       'img-src': ["'self'"],
       // Allow CSS loaded from https://fonts.googleapis.com
       'style-src': ["'self'", 'https://fonts.googleapis.com'],
+      // Allow manifest from the origin itself (i.e. current domain)
+      'manifest-src': ["'self'"],
       // Omit `media-src` from policy
       // Browser will fallback to default-src for media resources (which is 'none', see above)
       'media-src': null,


jelhan:
I'm wondering if we should update the default configuration as well: https://github.com/rwjblue/ember-cli-content-security-policy/blob/b6c3e093203acdba8b954442e49c0e91a6d25478/lib/utils.js#L73-L87 and https://github.com/rwjblue/ember-cli-content-security-policy/blob/b6c3e093203acdba8b954442e49c0e91a6d25478/README.md?plain=1#L94-L116

MichalBryxi:
Good point @jelhan. Added.

MichalBryxi:
Not sure why the CI gods are angry ...

jelhan:
> Not sure why the CI gods are angry ...

It seems to fail due to Prettier formatting:

```
$ eslint . --cache

/home/runner/work/ember-cli-content-security-policy/ember-cli-content-security-policy/lib/utils.js
ERROR: "lint:js" exited with 1.
Error:   78:7  error  Replace `·'default-src':·` with `'default-src':`                prettier/prettier
Error:   79:7  error  Replace `·'script-src':···` with `'script-src':·`               prettier/prettier
Error:   80:7  error  Replace `·'font-src':·····` with `'font-src':·`                 prettier/prettier
Error:   81:1  error  Replace `·······'connect-src':··` with `······'connect-src':·`  prettier/prettier
Error:   82:7  error  Replace `·'img-src':······` with `'img-src':·`                  prettier/prettier
Error:   83:7  error  Replace `·'style-src':····` with `'style-src':·`                prettier/prettier
Error:   84:1  error  Delete `·`                                                      prettier/prettier
Error:   85:7  error  Replace `·'media-src':···` with `'media-src':`                  prettier/prettier

✖ 8 problems (8 errors, 0 warnings)
  8 errors and 0 warnings potentially fixable with the `--fix` option.
```

MichalBryxi:
Even change in Readme would mean running CI, which is borked now. Just going to close this for now.

PR #366, opened 2024-04-25
chore: Bump node version used on GitHub

MichalBryxi:
- While [adding trivial change to the code](https://github.com/rwjblue/ember-cli-content-security-policy/pull/365) I bumped into CI being red from various reasons [last one](https://github.com/rwjblue/ember-cli-content-security-policy/actions/runs/8820844298/job/24215355798?pr=365) ended with a hard error on:
```
 [3/5] Fetching packages...
error ember-cli-babel@8.2.0: The engine "node" is incompatible with this module. Expected version "16.* || 18.* || >= 20". Got "12.22.12"
error Found incompatible module.
```
- I tried to reproduce locally, but `node@12.x` is surprisingly not trivial to install on modern OSX without breaking my other projects.
- I tried to consult the [node support](https://github.com/ember-cli/ember-cli/blob/master/docs/node-support.md) matrix for EmberJS and since `16.x` is supported through `3.28.0 - 5.3.0`, it seems like a safe bet.
- After a consultation in the PR, it has been decided to go with `node@18`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: '12.x'
+  NODE_VERSION: '18.x'
 
 jobs:
   lint:
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
-    "@ember/test-helpers": "^2.6.0",
+    "@ember/test-helpers": "^2.9.3",
     "@embroider/test-setup": "^0.49.0",
     "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.1.0",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@ember/test-helpers": "^2.9.3",
     "@embroider/test-setup": "^0.49.0",
     "@glimmer/tracking": "^1.1.2",
+    "@glimmer/manager": "^0.84.3",
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
     "chai": "^4.3.0",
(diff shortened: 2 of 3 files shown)


MichalBryxi:
Happy to hear your thoughts. I can't tell if this change _also_ means that the code compatibility will suffer? But some cuts need to be done at some point?

.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
-  NODE_VERSION: '12.x'
+  NODE_VERSION: '16.x'
```
jelhan: I'm happy to go with node 18 as node 16 reached end of life already.

jelhan (changes requested):
I'm happy dropping support for node <18. Could you please update `package.json` accordingly as well?

MichalBryxi:
> I'm happy dropping support for node <18. Could you please update `package.json` accordingly as well?

Should be done as well.

MichalBryxi:
It's "changing the lightbulb" again :) Now there seems to be an [old security issue](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported) in node(?) I tried to follow the most reasonably looking answer by [this commit](https://github.com/rwjblue/ember-cli-content-security-policy/pull/366/commits/7589bc055cf2ceac29fc217f572470422d9704ad).

Seems to easy to see that nothing dodgy is going on by:

```
npm_config_yes=true npx yarn-audit-fix
```

Now waiting for the CI to tell yes/no.

MichalBryxi:
This ... is getting pretty tangly. Any chance someone who might have a better overview over the project could take a look? I was trying to make a minimal change necessary, but ... things keep needing attention...

MichalBryxi:
Talked on Discord about this. The goal here is not to update the package per-se, but to allow me to progress with #365. Given that this package might get replaced, significant efforts to modernise it are not good use of anyone's time. So going to close this PR and see if #365 can be "fixed" simply by changing the README which _might_ pass the CI.

